PR #2, opened 2019-01-29
Add options to limit the number of allowed call timeouts

uartie:
Add --ctapr (call timeouts allowed per run).  If this limit
is reached in any single run, then all calls to the "call"
function will be rejected for all of the remaining tests.

Add --ctapt (call timeouts allowed per test function). Similar
to --ctapr, except the limit is only tracked on a per test function
scope.  For example, if test/ffmpeg-vaapi/decode/vp9.py:test_default
produces enough call timeouts to reach this limit, then all
remaining variations of this specific test function will not
succeed in a call to the "call" function.

--ctapr and --ctapt limits are only enforced in the scope after
test_start() and until test_end().


uartie:

PR #7, opened 2019-01-30
Only calculate checksum on num-frames

### LINKED ISSUE #3: md5 should only consider num-frames
Currently, md5 checksum metric is calculated on the whole file.   gstreamer decoding does not have the capability to only process N frames from an input file.  Therefore, it will always decode the entire input file which could have N+M frames.  This makes it hard to compare/use same checksum results between ffmpeg and gstreamer.

Rewrite the md5 checksum metric to only generate checksum from N frames.  This would also give the added benefit of detecting results that erroneously produce less-than N frames.

uartie:
Gstreamer does not have a way to limit the number
of decoded frames in the output.  Therefore, the
checksum was calculated on the entire decoded output
file and not on the number of frames that were
specified in the user configuration.  FFmpeg, on
the other hand, can output num-frames as requested
by the user configuration... and therefore, the
checksum is actually based on the number of frames.
This makes it difficult to easily compare between
all middleware to ensure they are properly producing
the same bitmatch decode output.

Change the md5 metric to only calculate the checksum
on the number of frames that were specified in the
user configuration.

Also, update gstreamer decode baselines accordingly.
Now checksum references are the same between the
ffmpeg and gstreamer decode cases, as they should be.

Fixes #3

### DIFF
--- a/lib/metrics.py
+++ b/lib/metrics.py
@@ -122,8 +122,8 @@ def calculate_psnr(filename1, filename2, width, height, nframes = 1, fourcc = "I
     sum(result[2::3]) / nframes,
   )
 
-def check_filesize(filename, width, height, nframes, fourcc):
-  expected = {
+def get_framesize(width, height, fourcc):
+  return {
     "I420" : lambda w,h: w*h*3/2,
     "422H" : lambda w,h: w*h*2,
     "422V" : lambda w,h: w*h*2,
@@ -134,10 +134,11 @@ def check_filesize(filename, width, height, nframes, fourcc):
     "Y800" : lambda w,h: w*h,
     "YUY2" : lambda w,h: w*h*2,
     "AYUV" : lambda w,h: w*h*4,
-  }[fourcc](width, height) * nframes
+  }[fourcc](width, height)
 
+def check_filesize(filename, width, height, nframes, fourcc):
+  expected = get_framesize(width, height, fourcc) * nframes
   actual = os.stat(filename).st_size
-
   assert expected == actual
 
 def check_metric(**params):
--- a/lib/metrics.py
+++ b/lib/metrics.py
@@ -12,11 +12,31 @@ import skimage.measure
 from common import get_media
 from framereader import FrameReaders
 
-def md5(filename):
+def md5(filename, chunksize = 4096, numbytes = -1):
+  if numbytes < 0: # calculate checksum on entire file
+    numbytes = os.stat(filename).st_size
+
+  numbytesread = 0
   m = hashlib.md5()
   with open(filename, "rb") as f:
-    for chunk in iter(lambda: f.read(4096), b""):
-      m.update(chunk)
+
+    # read numchunks of size chunksize
+    numchunks = int(numbytes / chunksize)
+    if numchunks > 0:
+      for n, chunk in enumerate(iter(lambda: f.read(chunksize), b""), 1):
+        numbytesread += len(chunk)
+        m.update(chunk)
+        if n == numchunks:
+          break
+
+    # read remainder of bytes < chunksize
+    lastchunk = f.read(numbytes % chunksize)
+    numbytesread += len(lastchunk)
+    m.update(lastchunk)
+
+  # fail if we did not read exactly numbytes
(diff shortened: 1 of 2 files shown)


uartie:

lib/metrics.py
```diff
@@ -12,11 +12,29 @@
...
+    numbytes = os.stat(filename).st_size
+
+  numbytesread = 0
   m = hashlib.md5()
   with open(filename, "rb") as f:
-    for chunk in iter(lambda: f.read(4096), b""):
+
+    # read numchunks of size chunksize
+    numchunks = int(numbytes / chunksize)
+    for n, chunk in enumerate(iter(lambda: f.read(chunksize), b""), 1):
```
wangzj0601: if numchunks = 0, it will calculate wrong, how about this.
for n, chunk in enumerate(iter(lambda: f.read(chunksize), b"")):
  if n == numchunks:
    break
  numbytesread += len(chunk)
  m.update(chunk)

uartie: You are right, we need to handle `numchunks = 0`... good catch.

However, with your suggested change, we would lose the first `chunksize` bytes.  That is, if `numbytes = 4095` and `chunksize = 4096`, then we still read `4096 bytes` from the file in the first loop iteration.  Thus, the `4095 bytes` read for `lastchunk` will be after the first `4096 bytes`.

What if we only do the loop if `numchunks > 0`? For example:
```
if numchunks > 0:
  for n, chunk in ...
lastchunk = ...
```

wangzj0601: if numbytes = 4095 and chunksize = 4096, then numchunks =0, it will break out the cycle because the n starts from 0 now, it is same with you.

wangzj0601: add the condition "if numchunks > 0:", it also looks good to me.

uartie: if `numchunks = 0`, we still enter the loop once while reading chunksize from the file.  I will update to check the `numchunks > 0` condition

uartie: patch updated.

wangzj0601: > if `numchunks = 0`, we still enter the loop once while reading chunksize from the file. I will update to check the `numchunks > 0` condition

Maybe the website display problem -_- -_- , I have put "break" condition before "reading chunksize from the file", it won't loop once when numchunks = 0.  -_- -_-

wangzj0601: > patch updated.

look good to me.

wangzj0601: > > if `numchunks = 0`, we still enter the loop once while reading chunksize from the file. I will update to check the `numchunks > 0` condition
> 
> Maybe the website display problem -_- -_- , I have put "break" condition before "reading chunksize from the file", it won't loop once when numchunks = 0. -_- -_-

sorry, you are right, it will does loop once, it is my mistake.

uartie: > > if `numchunks = 0`, we still enter the loop once while reading chunksize from the file. I will update to check the `numchunks > 0` condition
> 
> Maybe the website display problem -_- -_- , I have put "break" condition before "reading chunksize from the file", it won't loop once when numchunks = 0. -_- -_-

Ok, it could be a possible web rendering issue that made it hard to read.  You should style sample code between backticks, \`,  to get better rendering (https://help.github.com/articles/basic-writing-and-formatting-syntax/#quoting-code).

Thanks for reviewing!

wangzj0601: I dropped into the "for" cycle,  I suddenly thought that we said two different things, you said that it will loop once for reading file, I said that it won't loop once for "m.update(chunk)". So you are right, and it is right for adding the condition "if numchunks > 0:". It is my mistake, sorry for the noise.

lib/metrics.py
```diff
@@ -158,8 +179,13 @@ def check_metric(**params):
     assert 1.0 >= ssim[2] >= minv
 
   elif "md5" == type:
+    numbytes = get_framesize(
+      params["width"], params["height"],
```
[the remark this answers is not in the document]
uartie: vaapi-fits does not have multi-res case support yet.  Please file a feature request.

PR #8, opened 2019-01-31
validate test case names

uartie:
Ensure test case names conform to slash's printable values to avoid arbitrary and generic test name.


uartie:

PR #9, opened 2019-01-31
vpp/deinterlace: set interlace pipeline params

### LINKED ISSUE #4: deinterlace tests should set interlace parameters in pipelines
The deinterlace tests should assume all input files are interlaced content.  For gstreamer, we should set interlaced and top-field-first parameters for rawvideoparse.

We will need to investigate how to get ffmpeg to treat raw yuv as interlaced.  I am not sure how this is done.

uartie:
Refactor deinterlace tests to include interlace params
in the pipeline.  Also, set the default metric to use
md5 checksum (which can be configured in the user config).
The user config can set tff param:

 0 = bottom-field-first
 1 = top-field-first

By default, if tff is not set then 1 (top-field-first)
is assumed.

Fixes #4

### DIFF
--- a/test/ffmpeg-qsv/vpp/deinterlace.py
+++ b/test/ffmpeg-qsv/vpp/deinterlace.py
@@ -19,35 +19,27 @@ def test_default(case, method):
   params = spec[case].copy()
   params.update(
     method = map_deinterlace_method(method),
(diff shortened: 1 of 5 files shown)


uartie:

uartie:
@FocusLuo please help to review

PR #20, opened 2019-02-19
lib/metrics: calculate the md5 of whole file

wangzj0601:
if the case set a param "numbytes = -1", it can
calculate the md5 of whole file directly

### DIFF
--- a/lib/metrics.py
+++ b/lib/metrics.py
@@ -179,10 +179,10 @@ def check_metric(**params):
     assert 1.0 >= ssim[2] >= minv
 
   elif "md5" == type:
-    numbytes = get_framesize(
+    numbytes = metric.get("numbytes", get_framesize(
       params["width"], params["height"],
       params.get("format2", params["format"])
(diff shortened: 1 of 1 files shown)


lib/metrics.py
```diff
@@ -179,10 +179,10 @@ def check_metric(**params):
     assert 1.0 >= ssim[2] >= minv
 
   elif "md5" == type:
-    numbytes = get_framesize(
+    numbytes = params.get("numbytes", get_framesize(
```
uartie: `numbytes` should be retrieved from `metric = params["metric"]` sub-entry.

That is, the user config should specify in test case spec, like:
```
metric = dict(type = "md5", numbytes = -1)
```

Also, if user is allowed to set numbytes now, then it doesn't always have to be "whole file" mode... so please update commit message accordingly.  For example, "metrics: support user-specified md5 numbytes" and explain in the message body what behavior is expected based on user and default values.

uartie:
also, please rebase

Issue #19, opened 2019-02-16
list-config fails


djdeath:
```
$ ./vaapi-fits list-config
Extracting assets...
Traceback (most recent call last):
  File "./vaapi-fits", line 20, in <module>
    sys.exit(main())
  File "./vaapi-fits", line 17, in main
    return main_entry_point()
  File "/home/djdeath/.local/lib/python2.7/site-packages/slash/frontend/main.py", line 71, in main_entry_point
    sys.exit(main())
  File "/home/djdeath/.local/lib/python2.7/site-packages/slash/frontend/main.py", line 47, in main
    returned = func(args.argv)
  File "/home/djdeath/.local/lib/python2.7/site-packages/slash/frontend/list_config.py", line 29, in list_config
    slash.site.load()
  File "/home/djdeath/.local/lib/python2.7/site-packages/slash/site.py", line 17, in load
    return _load_defaults(working_directory=working_directory)
  File "/home/djdeath/.local/lib/python2.7/site-packages/slash/site.py", line 22, in _load_defaults
    _load_local_slashrc(working_directory=working_directory)
  File "/home/djdeath/.local/lib/python2.7/site-packages/slash/site.py", line 37, in _load_local_slashrc
    _load_file_if_exists(os.path.abspath(path))
  File "/home/djdeath/.local/lib/python2.7/site-packages/slash/site.py", line 44, in _load_file_if_exists
    _load_filename(path)
  File "/home/djdeath/.local/lib/python2.7/site-packages/slash/site.py", line 67, in _load_filename
    _load_source(f.read(), filename)
  File "/home/djdeath/.local/lib/python2.7/site-packages/slash/site.py", line 76, in _load_source
    exec(code, {"__file__" : filename}) # pylint: disable=W0122
  File "/home/djdeath/src/mesa-src/vaapi-fits/.slashrc", line 276, in <module>
    execfile(config)
  File "/home/djdeath/src/mesa-src/vaapi-fits/config/default", line 14, in <module>
    with tarfile.open("{}.tbz2".format(assets), "r:*") as tf:
  File "/usr/lib/python2.7/tarfile.py", line 1680, in open
    raise ReadError("file could not be opened successfully")
tarfile.ReadError: file could not be opened successfully
```

djdeath:
I think I just missed running `git lfs pull`

uartie:
Yes, we should provide a more user-friendly error message.

PR #21, opened 2019-02-19
config/default: improve assets error message

uartie:
Relates to #19

### DIFF
--- a/config/default
+++ b/config/default
@@ -11,8 +11,12 @@ assets = os.path.join(media.mypath, "assets")
 
 if not os.path.exists(assets):
   print "Extracting assets..."
-  with tarfile.open("{}.tbz2".format(assets), "r:*") as tf:
-    tf.extractall(media.mypath)
+  try:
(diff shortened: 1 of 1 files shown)


config/default
```diff
@@ -11,8 +11,12 @@ assets = os.path.join(media.mypath, "assets")
 
 if not os.path.exists(assets):
   print "Extracting assets..."
-  with tarfile.open("{}.tbz2".format(assets), "r:*") as tf:
-    tf.extractall(media.mypath)
+  try:
+    with tarfile.open("{}.tbz2".format(assets), "r:*") as tf:
+      tf.extractall(media.mypath)
+  except tarfile.ReadError, e:
```
wangzj0601: How about this?
except (tarfile.ReadError, IOError), e:

wangzj0601: if the file doesn't exists, we should also have a reminder

uartie: This patch is only meant to address the most common case that users may encounter.

PR #22, opened 2019-02-22
encode/mpeg2: map qp value into valid range

wangzj0601:
map qp to 2-62 for gst-vaapi
  map qp to 0-51 for gst-msdk

### DIFF
--- a/test/ffmpeg-vaapi/encode/encoder.py
+++ b/test/ffmpeg-vaapi/encode/encoder.py
@@ -29,8 +29,10 @@ class EncoderTest(slash.Test):
     if vars(self).get("gop", None) is not None:
       opts += " -g {gop}"
     if vars(self).get("qp", None) is not None:
-      if self.codec in ["mpeg2", "vp8", "vp9",]:
+      if self.codec in ["vp8", "vp9",]:
         opts += " -global_quality {qp}"
+      elif self.codec in ["mpeg2"]:
+        opts += " -global_quality {mqp}"
       else:
         opts += " -qp {qp}"
     if vars(self).get("quality", None) is not None:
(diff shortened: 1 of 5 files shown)


test/gst-msdk/encode/mpeg2.py
```diff
@@ -48,8 +48,8 @@ def test_cqp(case, gop, bframes, qp, quality, profile):
     slash.skip_test("{} profile is not supported".format(profile))
 
   params.update(
-    gop = gop, bframes = bframes, qp = qp, quality = quality,
-    profile = mprofile, mformat = mapformat(params["format"]))
+    gop = gop, bframes = bframes, qp = mapRange(qp, [0, 100], [0, 51]),
```
uartie: For modified/mapped variable names, please prefix with "`m`".  That is, use "`mqp`" instead of "`qp`" for the mapped qp result.  This makes it clearer that they are different.  We want to use the original/normalized value (`qp`) in, for example, the artifact filename.  But use the mapped value (`mqp`) in the pipeline command.

uartie:
If we want the user config to specify the mpeg2 enc QP in [0, 100] range, then we need to map the QP in the ffmpeg tests, too.

test/ffmpeg-vaapi/encode/encoder.py
```diff
@@ -100,6 +100,9 @@ def encode(self):
     if self.mformat is None:
       slash.skip_test("{format} format not supported".format(**vars(self)))
 
+    if self.codec in ["mpeg2"]:
+      self.qp = self.mqp
+
```
uartie: Please handle the mpeg2 mqp directly in `gen_output_opts` so we don't overwrite self.qp:

How about:
```
    if vars(self).get("qp", None) is not None:
      if self.codec in ["vp8", "vp9",]:
        opts += " -global_quality {qp}"
      elif self.codec in ["mpeg2"]:
        opts += " -global_quality {mqp}"
      else:
        opts += " -qp {qp}"
```

wangzj0601: have inlined the value into function "gen_output_opts", thanks.

PR #25, opened 2019-02-27
ffmpeg-vaapi/encode: add rc_mode option to pipeline

uartie:
Since ffmpeg commit:
 f4d7bd6cdbcf vaapi_encode: Support more RC modes

In ffmpeg, an "rc_mode" option was added to explicitly
set the rate-control mode.  If rc_mode is not specified,
then ffmpeg will try to auto-select the rc mode.
However, auto-select will choose AVBR over VBR if the
driver supports it.

Thus, set rc_mode explicitly to avoid auto-select.


lizhong1008:
LGTM

FocusLuo:
LGTM

PR #26, opened 2019-03-05
Csc optimize

wangzj0601:
support user-specified test standrad for csc and support format ARGB framereader.

### DIFF
--- a/lib/framereader.py
+++ b/lib/framereader.py
@@ -105,6 +105,17 @@ def read_frame_YUY2(fd, width, height):
 
   return y, u, v
 
+def read_frame_ARGB(fd, width, height):
+  size = width * height * 4
+
+  argb = numpy.fromfile(fd, dtype=numpy.uint8, count=size)
+  a = argb[0::4].reshape((height, width))
+  r = argb[1::4].reshape((height, width))
+  g = argb[2::4].reshape((height, width))
+  b = argb[3::4].reshape((height, width))
+
+  return r, g, b
+
 FrameReaders = {
   "I420" : read_frame_I420,
   "422H" : read_frame_422H,
@@ -116,4 +127,5 @@ FrameReaders = {
   "Y800" : read_frame_Y800,
   "YUY2" : read_frame_YUY2,
   "AYUV" : read_frame_AYUV,
+  "ARGB" : read_frame_ARGB,
 }
--- a/lib/metrics.py
+++ b/lib/metrics.py
@@ -154,6 +154,7 @@ def get_framesize(width, height, fourcc):
     "Y800" : lambda w,h: w*h,
     "YUY2" : lambda w,h: w*h*2,
     "AYUV" : lambda w,h: w*h*4,
+    "ARGB" : lambda w,h: w*h*4,
   }[fourcc](width, height)
 
 def check_filesize(filename, width, height, nframes, fourcc):
--- a/test/gst-msdk/vpp/csc.py
+++ b/test/gst-msdk/vpp/csc.py
@@ -1,0 +1,48 @@
+###
+### Copyright (C) 2018-2019 Intel Corporation
+###
+### SPDX-License-Identifier: BSD-3-Clause
+###
+
+from ....lib import *
+from ..util import *
+
+spec = load_test_spec("vpp", "csc")
+
+@slash.requires(have_gst)
+@slash.requires(*have_gst_element("msdk"))
+@slash.requires(*have_gst_element("msdkvpp"))
+@slash.requires(*have_gst_element("checksumsink2"))
+@slash.requires(using_compatible_driver)
+@slash.parametrize(*gen_vpp_csc_parameters(spec))
+@platform_tags(VPP_PLATFORMS)
+def test_default(case, csc):
+  params = spec[case].copy()
+  params.update(
+    csc = csc, mcscu = mapformatu(csc),
+    mformat = mapformat(params["format"]))
+  params["ofile"] = get_media()._test_artifact(
+    "{}_{format}_csc_{csc}_{width}x{height}"
+    ".yuv".format(case, **params))
+
+  call(
+    "gst-launch-1.0 -vf filesrc location={source} num-buffers={frames}"
+    " ! rawvideoparse format={mformat} width={width} height={height}"
+    " ! msdkvpp hardware=true ! video/x-raw,format={mcscu}"
+    " ! checksumsink2 file-checksum=false frame-checksum=false"
+    " plane-checksum=false dump-output=true dump-location={ofile}"
+    "".format(**params))
+
+  check_metric(
+    # if user specified metric, then use it.  Otherwise, use ssim metric with perfect score
+    metric = params.get("metric", dict(type = "ssim", miny = 1.0, minu = 1.0, minv = 1.0)),
+    # If user specified reference, use it.  Otherwise, assume source is the reference.
+    reference = format_value(params["reference"], case = case, **params)
+      if params.get("reference") else params["source"],
+    decoded = params["ofile"],
+    # if user specified reference, then assume it's format is the same as csc output format.
+    # Otherwise, the format is the source format
+    format = params["format"] if params.get("reference", None) is None else params["csc"],
+    format2 = params["csc"],
+    width = params["width"], height = params["height"], frames = params["frames"],
+  )
--- a/test/gst-vaapi/vpp/csc.py
+++ b/test/gst-vaapi/vpp/csc.py
@@ -30,11 +30,16 @@ def test_default(case, csc):
     " frame-checksum=false plane-checksum=false dump-output=true"
     " dump-location={ofile}".format(**params))
 
-  ssim = calculate_ssim(
-    params["source"], params["ofile"],
-    params["width"], params["height"], params["frames"],
-    params["format"], params["csc"])
-
-  get_media()._set_test_details(ssim = ssim)
-
-  assert sum(ssim) == 6.0, "Y, U and V values should not be changed by CSC"
+  check_metric(
+    # if user specified metric, then use it.  Otherwise, use ssim metric with perfect score
+    metric = params.get("metric", dict(type = "ssim", miny = 1.0, minu = 1.0, minv = 1.0)),
+    # If user specified reference, use it.  Otherwise, assume source is the reference.
+    reference = format_value(params["reference"], case = case, **params)
+      if params.get("reference") else params["source"],
+    decoded = params["ofile"],
+    # if user specified reference, then assume it's format is the same as csc output format.
+    # Otherwise, the format is the source format
+    format = params["format"] if params.get("reference", None) is None else params["csc"],
+    format2 = params["csc"],
+    width = params["width"], height = params["height"], frames = params["frames"],
+  )
--- a/test/gst-msdk/util.py
+++ b/test/gst-msdk/util.py
@@ -75,6 +75,7 @@ def mapformat(format):
     "P010"  : "p010-10le",
     "AYUV"  : "ayuv",
     "YUY2"  : "yuy2",
+    "ARGB"  : "argb",
   }.get(format, None)
 
 @memoize
@@ -86,6 +87,7 @@ def mapformatu(format):
     "P010"  : "P010_10LE",
     "AYUV"  : "AYUV",
     "YUY2"  : "YUY2",
+    "ARGB"  : "ARGB",
   }.get(format, None)
 
 @memoize
--- a/test/gst-vaapi/util.py
+++ b/test/gst-vaapi/util.py
@@ -99,6 +99,7 @@ def mapformat(format):
     "P010"  : "p010-10le",
     "AYUV"  : "ayuv",
     "YUY2"  : "yuy2",
+    "ARGB"  : "argb",
   }.get(format, None)
 
 @memoize
@@ -110,6 +111,7 @@ def mapformatu(format):
     "P010"  : "P010_10LE",
     "AYUV"  : "AYUV",
     "YUY2"  : "YUY2",
+    "ARGB"  : "ARGB",
   }.get(format, None)
 
 


wangzj0601:
can you have a reviw @uartie

test/gst-vaapi/vpp/csc.py
```diff
@@ -18,23 +18,20 @@
...
-    params["source"], params["ofile"],
-    params["width"], params["height"], params["frames"],
-    params["format"], params["csc"])
-
-  get_media()._set_test_details(ssim = ssim)
-
-  assert sum(ssim) == 6.0, "Y, U and V values should not be changed by CSC"
+  params.setdefault(
+    "metric", dict(type = "ssim", miny = 1.0, minu = 1.0, minv = 1.0))
+  check_metric(**params)
```
uartie: This won't work.  The `check_metric` function relies on `params["reference"]` and `params["decoded"]`.  Neither one are in this params dictionary.  You will need to change `params["ofile"]` to `params["decoded"]` and `params["reference"]` should be set to `params["source"]` if user has not already supplied it in the config (e.g. `params.setdefault("reference", params["source"])`)

lib/framereader.py
```diff
@@ -105,6 +105,25 @@ def read_frame_YUY2(fd, width, height):
...
+  g = argb[2::4].reshape((height, width))
+  b = argb[3::4].reshape((height, width))
+
+  #y = [0.299 * r[i] + 0.587 * g[i] + 0.114 * b[i] for i in range(nums)]
+  #u = [-0.147 * r[i] - 0.289 * g[i] + 0.436 * b[i] for i in range(nums)]
+  #v = [0.615 * r[i] - 0.515 * g[i]  - 0.100 * b[i] for i in range(nums)]
+
+  #y = numpy.array(y); u = numpy.array(u); v = numpy.array(v)
+
+  #return y.reshape((height, width)), u.reshape((height, width)), v.reshape((height, width))
```
uartie: Please remove commented code.

lib/metrics.py
```diff
@@ -170,7 +171,7 @@ def check_metric(**params):
     minu = metric.get("minu", 1.0)
     minv = metric.get("minv", 1.0)
     ssim = calculate_ssim(
-      params["reference"], params["decoded"],
+      params["reference"], params.get("decoded", params["ofile"]),
```
uartie: The test should set params["decoded"] instead.

lib/metrics.py
```diff
@@ -183,7 +184,7 @@ def check_metric(**params):
       params["width"], params["height"],
       params.get("format2", params["format"])
     ) * params["frames"])
-    res = md5(filename = params["decoded"], numbytes = numbytes)
+    res = md5(filename = params.get("decoded", params["ofile"]), numbytes = numbytes)
```
uartie: The test should set params["decoded"] instead.

test/gst-msdk/vpp/csc.py
```diff
@@ -0,0 +1,39 @@
...
+@slash.requires(*have_gst_element("msdkvpp"))
+@slash.requires(*have_gst_element("checksumsink2"))
+@slash.requires(using_compatible_driver)
+@slash.parametrize(*gen_vpp_csc_parameters(spec))
+@platform_tags(VPP_PLATFORMS)
+def test_default(case, csc):
+  params = spec[case].copy()
+  params.update(
+    csc = csc, mcscu = mapformatu(csc),
+    format2 = csc, mformat = mapformat(params["format"]))
```
uartie: format2 = csc is unnecessary

test/gst-msdk/vpp/csc.py
```diff
@@ -0,0 +1,39 @@
...
+  params["ofile"] = get_media()._test_artifact(
```
uartie: Use params["decoded"]

test/gst-msdk/vpp/csc.py
```diff
@@ -0,0 +1,39 @@
...
+    "{}_{format}_csc_{csc}_{width}x{height}"
+    ".yuv".format(case, **params))
+
+  call(
+    "gst-launch-1.0 -vf filesrc location={source} num-buffers={frames}"
+    " ! rawvideoparse format={mformat} width={width} height={height}"
+    " ! msdkvpp hardware=true ! video/x-raw,format=NV12"
+    " ! videoconvert ! video/x-raw,format={mcscu}"
```
uartie: This is doing software CSC via videoconvert.  The only thing being verified on msdkvpp here is conversion to NV12.  This is not an acceptable test.  The whole purpose is to test direct input->output CSC functionality in msdkvpp only.

uartie:
> can we use md5 as test standard?

I would prefer ssim be the default since it reveals more details about the result then md5

lib/framereader.py
wangzj0601: ok, I will remove it.

lib/metrics.py
wangzj0601: I think it is not accurate which all output are named decoded.

lib/metrics.py
wangzj0601: as above

test/gst-vaapi/vpp/csc.py
```diff
@@ -18,23 +18,20 @@
 def test_default(case, csc):
   params = spec[case].copy()
   params.update(
-    csc = csc, mcsc = mapformat(csc), mformat = mapformat(params["format"]))
+    csc = csc, mcsc = mapformat(csc), mformat = mapformat(params["format"]),
+    format2 = csc, hwup_format = mapformat_hwup(params["format"]))
```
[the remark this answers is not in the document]
wangzj0601: I set "format2 = csc" at this, because if we use function "check_metric(**params)", the variable will be used.

test/gst-vaapi/vpp/csc.py
```diff
@@ -18,23 +18,20 @@
...
   params["ofile"] = get_media()._test_artifact(
     "{}_{format}_csc_{csc}_{width}x{height}"
     ".yuv".format(case, **params))
 
   call(
     "gst-launch-1.0 -vf filesrc location={source} num-buffers={frames}"
     " ! rawvideoparse format={mformat} width={width} height={height}"
+    " ! videoconvert ! video/x-raw, format={hwup_format}"
```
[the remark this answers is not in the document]
wangzj0601: If we give an I420 input on iHD driver, the cmdline will fail, so what should we do?

wangzj0601:
> > can we use md5 as test standard?
> 
> I prefer ssim be the default since it reveals more details about the result than md5

If we use ssim as standard, we should create reference for every case, because "assert sum(ssim) == 6.0" only can be used when both input and output have the same pixel subsampling, then we do ssim during output and reference, do you agree this solution?

test/gst-msdk/vpp/csc.py
wangzj0601: ok, I will remove it. Alos, if we give an I420 input on iHD driver, the cmdline will fail, so what should we do?

wangzj0601:
> > can we use md5 as test standard?
> 
> I prefer ssim be the default since it reveals more details about the result than md5

Of course, we should confirm the test standard firstly, then I can modify my code better.

test/gst-vaapi/vpp/csc.py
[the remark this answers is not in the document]
uartie: "format2" is only used if it is defined in params.  In check_metric, params.get("format2", None) for ssim metric and params.get("format2", params["format"]) for md5 metric handle this.  Please brush up on Python dict type documentation.

test/gst-vaapi/vpp/csc.py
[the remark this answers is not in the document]
uartie: Failure in that case is expected.  The intention of this CSC test is to test the expected features of the vaapi driver csc support from format A to format B without doing any special, intermediate conversions.  If iHD driver states they will ***never*** support this type of conversion, then such a case should not be defined in our test configuration.

lib/metrics.py
uartie: If you have a better, more "generic" name, then please feel free to propose it.

lib/metrics.py
uartie: as above

test/gst-msdk/vpp/csc.py
uartie: as above

uartie:
> > > can we use md5 as test standard?
> > 
> > 
> > I prefer ssim be the default since it reveals more details about the result than md5
> 
> If we use ssim as standard, we should create reference for every case, because "assert sum(ssim) == 6.0" only can be used when both input and output have the same pixel subsampling, then we do ssim during output and reference, do you agree this solution?

That is why it should be user configurable.  User configurable metric support wasn't added until later on... and only some tests got converted to allow configurable metric.  We just need to finish updating the rest of the tests.

test/gst-vaapi/vpp/csc.py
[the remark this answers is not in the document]
uartie: Oh, sorry.  You are right.  I forgot you need format2 defined for csc test.

lib/metrics.py
uartie: It doesn't make sense for check_metric to handle all possible output names that a test could use.  Thus, a standard/generic name is easier to maintain at this level.

test/gst-msdk/vpp/csc.py
uartie: Oh, sorry. You are right. I forgot you need format2 defined for csc test.

uartie:
> > > can we use md5 as test standard?
> > 
> > 
> > I prefer ssim be the default since it reveals more details about the result than md5
> 
> Of course, we should confirm the test standard firstly, then I can modify my code better.

This csc test was originally written to only support CSC with same subsampling, thus ssim works fine for that.  But of course, making the metric configurable (as stated before) will allow us to define cases with different subsampling and reference files if user wants.  There should not be any problem adding that functionality while still retaining the current behavior as the default.  Not all features are available in this entire test suite that we want... thus I am grateful for any enhancements provided.

wangzj0601:
Do you mean that we still that use "sum(ssim) == 6.0" for same subsampling, and use reference file to do ssim for different subsampling? If so, I will modify my code and add condition to fit it.

uartie:
> Do you mean that we still that use "sum(ssim) == 6.0" for same subsampling, and use reference file to do ssim for different subsampling? If so, I will modify my code and add condition to fit it.

I think we can use ssim default like the fixed decode tests (e.g. avc) and user reference if specified else source  (e.g. `params.setdefault("reference", params["source"])`)

uartie:
> > Do you mean that we still that use "sum(ssim) == 6.0" for same subsampling, and use reference file to do ssim for different subsampling? If so, I will modify my code and add condition to fit it.
> 
> I think we can use ssim default like the fixed decode tests (e.g. avc) and user reference if specified else source (e.g. `params.setdefault("reference", params["source"])`)

And if user specifies a reference file, then assume it is in the output format (csc).

wangzj0601:
OK, I see, I modify it later, thank you very much.

test/gst-msdk/vpp/csc.py
```diff
@@ -0,0 +1,54 @@
...
+  else:
+    ssim = calculate_ssim(
+    params["reference"], params["ofile"],
+    params["width"], params["height"], params["frames"],
+    params["csc"])
+
+    get_media()._set_test_details(ssim = ssim)
+    assert 1.0 >= ssim[0] >= 0.99
+    assert 1.0 >= ssim[1] >= 0.99
+    assert 1.0 >= ssim[2] >= 0.99
```
uartie: I thought we decided this should be user configurable (i.e. the metric).  Let the user set the reference file.  The test should not need to check if subsampling is the same or not.  It should be up to the user to set the reference correctly (i.e. one with same subsample as the csc output).

test/gst-msdk/vpp/csc.py
```diff
@@ -0,0 +1,54 @@
...
+    csc = csc, mcscu = mapformatu(csc),
+    mformat = mapformat(params["format"]))
+  params["ofile"] = get_media()._test_artifact(
+    "{}_{format}_csc_{csc}_{width}x{height}"
+    ".yuv".format(case, **params))
+
+  call(
+    "gst-launch-1.0 -vf filesrc location={source} num-buffers={frames}"
+    " ! rawvideoparse format={mformat} width={width} height={height}"
+    " ! msdkvpp hardware=true ! videoconvert ! video/x-raw,format={mcscu}"
```
uartie: This is still wrong.  msdkvpp is doing nothing here.  CSC is being done by videoconvert which is SW CSC.  You need to do CSC with msdkvpp

test/gst-msdk/vpp/csc.py
uartie: Thus, this should work:
```python
check_metric(
  # if user specified metric, then use it.  Otherwise, use ssim metric with perfect score
  metric = params.get("metric", dict(type = "ssim", miny = 1.0, minu = 1.0, minv = 1.0)),
  # If user specified reference, use it.  Otherwise, assume source is the reference.
  reference = params.get("reference", params["source"]),
  decoded = params["ofile"],
  # if user specified reference, then assume it's format is the same as csc output format.
  # Otherwise, the format is the source format
  format = params["format"] if params.get("reference", None) is None else params["csc"],
  format2 = params["csc"],
  width = params["width"], height = params["height"], frames = params["frames"],
)

Comment (wangzj0601) [reply test/gst-msdk/vpp/csc.py]: During modifying the code, I thought a potential problem which the user don't set reference and input/output are different subsampling, then the result will fail, so I add the condition again.
How about
```
if same_pixel_subsample(params["format"], params["csc"]) and (not params.get("reference")):
  params["source"], params["ofile"]
else:
  check_metric(
    metric = params.get("metric", dict(type = "ssim", miny = 1.0, minu = 1.0, minv = 1.0)),
    decoded = params["ofile"],
    format = params["csc"],
    width = params["width"], height = params["height"], frames = params["frames"],
)
```

File: test/gst-msdk/vpp/csc.py (hunk 11 above)
Comment (wangzj0601) [reply test/gst-msdk/vpp/csc.py]: I follow the full test spec to design this, I will investigate it.

Comment (uartie) [reply test/gst-msdk/vpp/csc.py]: The test spec is probably wrong.  @FocusLuo any comment?

File: test/gst-msdk/vpp/csc.py (hunk 10 above)
Comment (uartie) [reply test/gst-msdk/vpp/csc.py]: > During modifying the code, I thought a potential problem which the user don't set reference and input/output are different subsampling, then the result will fail, so I add the condition again.

Then that is user error and clearly should go back and define the test case properly.

> How about
> ```
> if same_pixel_subsample(params["format"], params["csc"]) and (not params.get("reference")):
>   check_metric(
>     metric = params.get("metric", dict(type = "ssim", miny = 1.0, minu = 1.0, minv = 1.0)),
>     decoded = params["ofile"],
>     reference = params["source"],
>     format = params["format"],
>     format2 = params["csc"],
>     width = params["width"], height = params["height"], frames = params["frames"],
>   )
> else:
>   check_metric(
>     metric = params.get("metric", dict(type = "ssim", miny = 1.0, minu = 1.0, minv = 1.0)),
>     decoded = params["ofile"],
>     reference = params["reference"],
>     format = params["csc"],
>     width = params["width"], height = params["height"], frames = params["frames"],
>   )
> ```
> should be like this

I think this is unnecessary complexity.  We don't need same_pixel_subsample since it is up to the user to do define the case properly.  Please go with my previous solution.

Comment (uartie) [reply test/gst-msdk/vpp/csc.py]: It's not like failing cases will go unnoticed.

File: test/gst-msdk/vpp/csc.py [hunk 12]
```diff
@@ -0,0 +1,48 @@
...
+    mformat = mapformat(params["format"]))
+  params["ofile"] = get_media()._test_artifact(
+    "{}_{format}_csc_{csc}_{width}x{height}"
+    ".yuv".format(case, **params))
+
+  call(
+    "gst-launch-1.0 -vf filesrc location={source} num-buffers={frames}"
+    " ! rawvideoparse format={mformat} width={width} height={height}"
+    " ! msdkvpp hardware=true ! video/x-raw,format={mcscu}"
+    " ! checksumsink2 file-checksum=false frame-checksum=false"
```
uartie: checksumsink2 only supports `format: { (string)NV12, (string)I420, (string)YV12, (string)AYUV, (string)YUY2, (string)P010_10LE }`.  Please submit a patch to https://github.com/ceyusa/gst-checksumsink to add ARGB support.

wangzj0601: please refer to https://github.com/ceyusa/gst-checksumsink/pull/4

PR #27, opened 2019-03-08
test/ffmpeg-vaapi: do not use requires at class level

uartie:
Using slash.requires at the class level causes the
requirement to be applied to all branches of the
entire class inheritance tree starting from the base
class.  This is problematic because if one requirement
is not met on a derived class, then every class that
is derived from the same base gets flagged for that
unmet requirement.

See https://github.com/getslash/slash/issues/928 for
more details about this issue.

Thus, move slash.requires to the test functions where
we intend for that requirement only to be used within
that scope.

Also, to be safe, don't use slash.tags at the class
level either.


uartie:
thanks for reviewing

PR #33, opened 2019-03-13
gst-vaapi/gst-msdk: add decode scaling option

uartie:
Add option "scale_output" to allow the test case config
to instruct the gst-vaapi or gst-msdk decode tests to
apply videoscale (SW scaling) to the output.  This is
mostly useful for testing decode cases where the input
file has multiple resolutions and we want decoded output
to be in a single resolution so that it can work with
certain metrics settings.

By default, ffmpeg automatically scales (SW scaling)
output (which can't be disabled, afaik).  Thus, ffmpeg
tests don't need to do anything with scale_output.

### DIFF
--- a/test/gst-msdk/decode/decoder.py
+++ b/test/gst-msdk/decode/decoder.py
@@ -23,10 +23,16 @@ class DecoderTest(slash.Test):
     self.decoded = get_media()._test_artifact(
       "{case}_{width}x{height}_{format}.yuv".format(**vars(self)))
 
+    self.gstscaler = ""
+    if vars(self).get("scale_output", False):
+      self.gstscaler = (
+        " ! videoscale"
+        " ! video/x-raw,width={width},height={height}".format(**vars(self)))
+
     call(
       "gst-launch-1.0 -vf filesrc location={source}"
       " ! {gstdecoder} ! videoconvert ! video/x-raw,format={mformatu}"
-      " ! checksumsink2 file-checksum=false qos=false"
+      " {gstscaler} ! checksumsink2 file-checksum=false qos=false"
       " frame-checksum=false plane-checksum=false dump-output=true"
       " dump-location={decoded}".format(**vars(self)))
 
--- a/test/gst-vaapi/decode/decoder.py
+++ b/test/gst-vaapi/decode/decoder.py
@@ -22,10 +22,16 @@ class DecoderTest(slash.Test):
     self.decoded = get_media()._test_artifact(
       "{case}_{width}x{height}_{format}.yuv".format(**vars(self)))
 
+    self.gstscaler = ""
+    if vars(self).get("scale_output", False):
+      self.gstscaler = (
+        " ! videoscale"
+        " ! video/x-raw,width={width},height={height}".format(**vars(self)))
+
     call(
       "gst-launch-1.0 -vf filesrc location={source}"
       " ! {gstdecoder}"
-      " ! videoconvert ! video/x-raw,format={mformatu}"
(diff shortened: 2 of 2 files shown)


xhaihao:
What is the criteria after using videoscale?

uartie:
The criteria/metric is configurable per case... either MD5 checksum or SSIM reference.  SSIM minimum acceptable is also configurable.  By default, the metric is SSIM (1.0 min for fixed and 0.99 min for float).  The default makes sense for most standard cases (e.g. single resolution).  However, for multi resolution cases I think we'll configure it to use MD5 checksum ref per middleware.

uartie:
Here is an example of how we might specify a multi-resolution case in our test suite configuration file:
```
media._get_test_spec("avc", "decode").update({
 "multi-res-case-name" : dict(
    source = os.path.join(assets, "avc", "multi-res-case.h264"),
    scale_output = True, metric = dict(type = "md5"),
    format = "NV12", width = 352, height = 288, frames = 100),
})
```
Each middleware test (i.e. ffmpeg-vaapi, ffmpeg-qsv, gst-vaapi, gst-msdk) will load this case and execute according to these settings.  That is, decode `100 frames` from the `source` and `scale` all frames of the output to `width` x `height` in `NV12 format`.  Each test will lookup the `md5` reference value (from the baseline file) using the middleware name + test case name as the key and compare it with the `md5` result of the output.

Additionally, we could also specify that the reference be further indexed on a per-driver basis by defining `refctx = ["driver"]` in the test case config.  Thus, middleware name + test case name + driver would be the lookup key instead.

xhaihao:
How do you get the md5 reference value? compared to software decoder + videoscale?

xhaihao:
Is there a way to narrow down whether the issue is caused by vidoscale when md5 value is mismatched with reference value?

uartie:
> How do you get the md5 reference value? compared to software decoder + videoscale?

In general, vaapi-fits isn't concerned with how references are obtained and their correctness.

This is a detail left to the person defining the test case and its initial reference.  If they choose, they can simply use the first result from the test suite run and save that as the reference (e.g. `--rebase` option). 
 Or any reference can be added to the baseline file manually.  Obviously, there is no guarantee that the output used to generate a reference is always correct when using these practices.  Thus, further measures and analysis should be done to ensure correct reference values before using them as a "golden" reference.   You can consult the QA team to discover what their methodology is to set "golden" references.

uartie:
> Is there a way to narrow down whether the issue is caused by vidoscale when md5 value is mismatched with reference value?

Sure, there are many ways.  Just be creative.  And vaapi-fits has a lot of useful options to help a user with that process.

xhaihao:
Thanks Artie, I'm fine to add this option.

lizhong1008:
@uartie , is it works well for ffmpeg-qsv auto-scaling? IIRC, it will failed if decoding is working on GPU memory, see https://github.com/Intel-FFmpeg-Plugin/Intel_FFmpeg_plugins/issues/48#issuecomment-436541792

uartie:
> @uartie , dose it work well for ffmpeg-qsv with auto-scaling? IIRC, it will failed if decoding is working on GPU memory, see [Intel-FFmpeg-Plugin/Intel_FFmpeg_plugins#48 (comment)](https://github.com/Intel-FFmpeg-Plugin/Intel_FFmpeg_plugins/issues/48#issuecomment-436541792)

This patch does not change how the current ffmpeg tests behave.  Thus, I think that issue is out of scope for this pull-request.  We should follow-up on that particular issue in another thread if needed.

PR #36, opened 2019-03-25
 Default test timeout time increased

haribommi:
Default test timeout time increased from 300 seconds to 600 seconds for allowing larger files to complete decoding.
Ex: VP9 10-bit 4K test stream TheWorldinHDR.mkv taking around 600 seconds to complete VP9 decoding.

Signed-off-by: hbomminx <haribabux.bommineni@intel.com>

### DIFF
--- a/.slashrc
+++ b/.slashrc
@@ -64,7 +64,7 @@ class MediaPlugin(slash.plugins.PluginInterface):
       type = int,
       help = "{} = Keep None; {} = Keep Failed; {} = Keep All".format(
         self.RETENTION_NONE, self.RETENTION_FAIL, self.RETENTION_ALL))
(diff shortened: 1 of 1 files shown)


haribommi:
@uartie , @FocusLuo and @wangzj0601  couls you please review this patch

haribommi:
gst-vaapi i965 taking longer time than ffmpeg-vaapi decoding for container test stream

uartie:
Any reason you can't use `--call-timeout` option?

haribommi:
yes, with --call-timeout 720 also tests are passing but thought changing here will work for larger test streams.

uartie:
I think setting to 720 by default is too long for most cases.  Maybe it would be better to add support in user config to specify a call timeout on a per-case basis.

PR #45, opened 2019-04-16
test: enable hevc 422 and 444 10bit decoding

wangzj0601:


wangzj0601:
depends on https://github.com/ceyusa/gst-checksumsink/pull/5

test/ffmpeg-qsv/decode/10bit/hevc.py
```diff
@@ -9,6 +9,8 @@
 from ..decoder import DecoderTest
 
 spec = load_test_spec("hevc", "decode", "10bit")
+spec_422 = load_test_spec("hevc", "decode", "10bit", "422p")
+spec_444 = load_test_spec("hevc", "decode", "10bit", "444p")
```
uartie: What is 422p and 444p?  Doesn't "p" indicate "planar"?  How do these names have any relation to Y210 and Y410?

lib/platform.py
```diff
@@ -53,28 +55,30 @@
...
+  AVC_DECODE_PLATFORMS            = [                   "BDW",        "SKL", "APL", "KBL",        "CFL", "WHL", "ICL"]
+  AVC_ENCODE_PLATFORMS            = [                   "BDW",        "SKL", "APL", "KBL",        "CFL", "WHL", "ICL"]
+  AVC_ENCODE_LP_PLATFORMS         = [                                 "SKL", "APL", "KBL",        "CFL", "WHL", "ICL"]
+  HEVC_DECODE_8BIT_PLATFORMS      = [                                 "SKL", "APL", "KBL",        "CFL", "WHL", "ICL"]
+  HEVC_DECODE_8BIT_8K_PLATFORMS   = [                                 "SKL", "APL", "KBL",        "CFL", "WHL", "ICL"]
+  HEVC_ENCODE_8BIT_PLATFORMS      = [                                 "SKL", "APL", "KBL",        "CFL", "WHL", "ICL"]
+  HEVC_ENCODE_8BIT_LP_PLATFORMS   = [                                                                           "ICL"]
+  HEVC_DECODE_10BIT_PLATFORMS     = [                                        "APL", "KBL",        "CFL", "WHL", "ICL"]
+  HEVC_DECODE_10BIT_422_PLATFORMS = [                                                                           "ICL"]
+  HEVC_DECODE_10BIT_444_PLATFORMS = [                                                                           "ICL"]
```
uartie: Unfortunately, the original idea was to use these only for `codec->platform` mappings.  Now we're starting to see `resolution->platform` and `pix format->platform` mappings creep into this design.  This is making it worse to maintain (i.e. it forces us to add more specialized tags, test functions and user config keys).  Maybe we should start thinking about a better way to handle these latter cases at test runtime.  We don't have to do it now, but let's think about it later.  I have a few ideas.

Your current solution is ok for now.

test/ffmpeg-vaapi/decode/10bit/hevc.py
```diff
@@ -9,6 +9,8 @@
 from ..decoder import DecoderTest
 
 spec = load_test_spec("hevc", "decode", "10bit")
+spec_422 = load_test_spec("hevc", "decode", "10bit", "422p")
+spec_444 = load_test_spec("hevc", "decode", "10bit", "444p")
```
uartie: Maybe it would be better not to introduce a new spec key.  Rather, we could just filter the original/existing spec based on format value during tests `@slash.parametrize` call.  For example:

```
@slash.parametrize(("case"), sorted([k for k,v in spec.items() if v["format"] in ["P010"]]))
```

lib/platform.py
wangzj0601: how about this
DECODE_10BIT_422_PLATFORMS = [ ... ]
@platform_tags(set(HEVC_DECODE_10BIT_PLATFORMS) & set(DECODE_10BIT_422_PLATFORMS))

test/ffmpeg-vaapi/decode/10bit/hevc.py
wangzj0601: have change it

test/ffmpeg-qsv/decode/10bit/hevc.py
wangzj0601: remove this

lib/platform.py
uartie: yes, that is better.

test/gst-msdk/util.py
```diff
@@ -88,6 +92,10 @@ def mapformatu(format):
     "AYUV"  : "AYUV",
     "YUY2"  : "YUY2",
     "ARGB"  : "ARGB",
+    "422H"  : "Y42B",
+    "444P"  : "Y444",
```
uartie: Please just focus on the 422/444 10bit changes in this pull-request.

uartie:
I see that iHD driver has a patch for Y210/Y410 (i.e. packed 422/444 10 bit).  However, this patch is for P210/P410 test enabling.  Does iHD plan to support P210/P410 (i.e. planar 422/444 10 bit)?

test/ffmpeg-qsv/util.py
```diff
@@ -102,6 +102,16 @@ def mapformat(format):
...
+    "P210"  : "yuv422p10le",
+    "P410"  : "yuv444p10le",
+  }.get(format, None)
+
+@memoize
+def mapformathw(format):
+  return {
+    "P010"  : "p010le",
+    "P210"  : "p210le",
+    "P410"  : "p410le",
```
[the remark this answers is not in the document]
wangzj0601: Yes,  p210le and p410le do not support in ffmpeg now, but after sync with ffmpeg developer, they agree these name, and they will use these name for dev, so we can use them in advance and if they have some little change in the future, we can give a patch to fix it.

test/gst-msdk/util.py
wangzj0601: ok, I will revert the commit for 422/444 8bit decoding

wangzj0601:
> I see that iHD driver has a patch for Y210/Y410 (i.e. packed 422/444 10 bit). However, this patch is for P210/P410 test enabling. Does iHD plan to support P210/P410 (i.e. planar 422/444 10 bit)?

Yes, iHD driver only supports Y210/Y410 now, but we can still use P210/P410 for output format because the middleware can convert the format to P210/P410 via SW method in our cmdline. 
In addition, I only have P210/P410 golden reference md5, so I want to use P210/P410 as output format

test/ffmpeg-qsv/decode/hevc.py
```diff
@@ -18,7 +18,8 @@ def before(self):
 
   @platform_tags(HEVC_DECODE_8BIT_PLATFORMS)
   @slash.requires(have_ffmpeg_hevc_qsv_decode)
-  @slash.parametrize(("case"), sorted([k for k,v in spec.items() if v["width"] <= 4096 and v["height"] <= 4096]))
+  @slash.parametrize(("case"), sorted([k for k,v in spec.items() if v["width"] <= 4096
+    and v["height"] <= 4096 and v["format"] in ["I420"]]))
```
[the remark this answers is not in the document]
wangzj0601: use vainfo on on ICL, you will find below info, but not on KBL, so I think it still need to specialize 8bit format. Howerver, as you said,  let's only focus on the 422/444 10bit firstly now.
VAProfileHEVCMain422_10         : VAEntrypointVLD
VAProfileHEVCMain444            : VAEntrypointVLD

test/ffmpeg-qsv/util.py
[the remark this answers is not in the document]
uartie: If ffmpeg is going to support p210le/p410le as synonym for yuv422p10le/yuv444p10le, then we only need to use one of them.  Thus, mapformatsw is unnecessary.

uartie:
> > I see that iHD driver has a patch for Y210/Y410 (i.e. packed 422/444 10 bit). However, this patch is for P210/P410 test enabling. Does iHD plan to support P210/P410 (i.e. planar 422/444 10 bit)?
> 
> Yes, iHD driver only supports Y210/Y410 now, but we can still use P210/P410 for output format because the middleware can convert the format to P210/P410 via SW method in our cmdline.
> In addition, I only have P210/P410 golden reference md5, so I want to use P210/P410 as output format

Yes, but you are using the unsupported driver formats (p210le/p410le) in the hwformat variable in the tests.  This variable specifies the "hwdownload,format={hwformat}" pipeline parameter.  And hwdownload requires the driver to have support for that format.

wangzj0601:
> > > I see that iHD driver has a patch for Y210/Y410 (i.e. packed 422/444 10 bit). However, this patch is for P210/P410 test enabling. Does iHD plan to support P210/P410 (i.e. planar 422/444 10 bit)?
> > 
> > 
> > Yes, iHD driver only supports Y210/Y410 now, but we can still use P210/P410 for output format because the middleware can convert the format to P210/P410 via SW method in our cmdline.
> > In addition, I only have P210/P410 golden reference md5, so I want to use P210/P410 as output format
> 
> Yes, but you are using the unsupported driver formats (p210le/p410le) in the hwformat variable in the tests. This variable specifies the "hwdownload,format={hwformat}" pipeline parameter. And hwdownload requires the driver to have support for that format.

Yes, you are right, the driver must support the "hwdownload" format, how about "'hwdownload,format=y210le' -pix_fmt yuv422p10le" and "'hwdownload,format=y410le' -pix_fmt yuv444p10le", can this will work?

uartie:
> Yes, you are right, the driver must support the "hwdownload" format, how about "'hwdownload,format=y210le' -pix_fmt yuv422p10le" and "'hwdownload,format=y410le' -pix_fmt yuv444p10le", can this will work?(BTW, the y210le/y410le are also synced with ffmpeg developer)

Yes, I think that would work fine without losing pixel data.

test/ffmpeg-qsv/util.py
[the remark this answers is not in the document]
wangzj0601: remove the fun "mapformatsw"

PR #57, opened 2019-04-26
Transcode scale

haribommi:


haribommi:
user-spec example:
 "H_H-HEVC-4K-HEVC-2K-AVC-1080P-MPEG2-1080P-AVC-720P-MJPEG-480P-HEVC-D1-AVC-CIF-cabac" : dict(
    source = os.path.join(assets, "avc", "720x480p_30.00_4mb_h264_cabac_180s.264"),
    width = 720, height = 480, frames = 200, mode = "hw",
    outputs = [
      dict(codec = "hevc", mode = "hw", width = 4096, height = 2304),
      dict(codec = "hevc", mode = "hw", width = 2048, height = 1080),
      dict(codec = "avc", mode = "hw", width = 1920, height = 1080),
      dict(codec = "mpeg2", mode = "hw", width = 1920, height = 1080),
      dict(codec = "avc", mode = "hw", width = 1280, height = 720),
      dict(codec = "mjpeg", mode = "hw", width = 720, height = 480),
      dict(codec = "hevc", mode = "hw", width = 704, height = 576),
      dict(codec = "avc", mode = "hw", width = 352, height = 288),
    ]),

test/ffmpeg-qsv/transcode/transcoder.py
```diff
@@ -76,6 +79,13 @@ def validate_spec(self):
       codec = output["codec"]
       mode  = output["mode"]
       assert mode in ["sw", "hw"], "Invalid test case specification as output mode type not valid"
+      width = output.get("width", 0)
+      height = output.get("height", 0)
+
+      if width > 0 and height > 0:
+        if not (self.mode == "hw" and mode == "hw"):
+          slash.skip_test("incorrect mode, scaling not enabled")
```
uartie: This should be a test failure (i.e. `assert`) since it is a user spec violation.

But this also begs the question... why can't we support sw scaling?

test/ffmpeg-qsv/transcode/transcoder.py
```diff
@@ -154,6 +169,39 @@ def check_output(self):
...
     assert m is None, "Failed to use hardware decode"
 
+  # scale scaled test stream to source file width and height for generating yuv
+  def scale_to_source_size(self, scaled, codec, mode, width, height, frames, enc_file):
+    _, _, ffdecparms = self.get_requirements_data("decode", codec, mode)
+    ffdeccodec, _ = ffdecparms
+    _, _, ffencparms = self.get_requirements_data("encode", codec, mode)
+    ffenccodec, ext = ffencparms
+    # TBD - for mjpeg and mpeg2, HW scaling not working so using SW scaling
+    # Remove this once HW scaling works for jpeg and mjpeg2
```
uartie: If it's not working, then it's a middleware bug... so why would we want to hide this bug?  Please don't hide the bug.

test/ffmpeg-qsv/transcode/transcoder.py
```diff
@@ -154,6 +169,39 @@ def check_output(self):
...
+      opts += " -vf scale=w={}:h={}".format(width, height)
+    elif ext == "m2v":
+      opts += " -c:v mpeg2video"
+      opts += " -vf scale=w={}:h={}".format(width, height)
+    else:
+      opts += " -c:v {}".format(ffenccodec)
+      opts += " -vf vpp_qsv=w={}:h={}".format(width, height)
+
+    opts += " -vframes {}".format(frames)
+    encoded = enc_file + "_ref.{}".format(ext)
```
uartie: All artifacts should be registered with `get_media().test_artifact()` so it is properly managed.  It looks like `enc_file` is registered before calling this method.  But `enc_file + "_ref.{}` is not.

test/ffmpeg-qsv/transcode/transcoder.py
```diff
@@ -173,7 +221,17 @@ def transcode(self):
...
+        height = output.get("height", 0)
+
+        if width > 0 and height > 0:
+          scaled = self.goutputs[n][channel]
+          enc_file = get_media()._test_artifact(
+            "{}_{}_{}".format(self.case, n, channel))
+          encoded = self.scale_to_source_size(scaled, output["codec"], self.mode, self.width, self.height, self.frames, enc_file)
+        else:
+          encoded = self.goutputs[n][channel]
+
```
uartie: We should purge the encoded file too.

test/ffmpeg-qsv/transcode/transcoder.py
```diff
@@ -154,6 +169,39 @@ def check_output(self):
...
+    ffenccodec, ext = ffencparms
+    # TBD - for mjpeg and mpeg2, HW scaling not working so using SW scaling
+    # Remove this once HW scaling works for jpeg and mjpeg2
+    if ext == "mjpeg":
+      opts = " -c:v mjpeg"
+    elif ext == "m2v":
+      opts = " -c:v mpeg2video"
+    else:
+      opts = " -hwaccel qsv -hwaccel_device /dev/dri/renderD128 -init_hw_device qsv=hw -c:v {} ".format(ffdeccodec)
+
```
uartie: You decode+scale to coded file in one call, then decode (the scaled coded file) to yuv in another call.  This is unnecessary.  It would be better to decode+scale directly to yuv in a single call.

test/ffmpeg-vaapi/transcode/transcoder.py
```diff
@@ -27,6 +27,9 @@ class TranscoderTest(slash.Test):
       "mpeg2" : dict(
         hw = (MPEG2_DECODE_PLATFORMS, (True, True), None),
       ),
+      "mjpeg" : dict(
+        hw = (JPEG_DECODE_PLATFORMS, (True, True), None),
+      ),
       "vc1" : dict(
         hw = (VC1_DECODE_PLATFORMS, (True, True), None),
       ),
```
uartie: What about vpp constraints/requirements?

test/ffmpeg-vaapi/transcode/transcoder.py
```diff
@@ -170,7 +201,17 @@ def transcode(self):
...
       for channel in xrange(output.get("channels", 1)):
-        encoded = self.goutputs[n][channel]
+        width = output.get("width", 0)
+        height = output.get("height", 0)
+
+        if width > 0 and height > 0:
+          scaled = self.goutputs[n][channel]
+          enc_file = get_media()._test_artifact(
+            "{}_{}_{}".format(self.case, n, channel))
+          encoded = self.scale_to_source_size(scaled, output["codec"], self.mode, self.width, self.height, self.frames, enc_file)
```
uartie: There is no need to pass `self.*` parameters to member function since you can directly access them from the member function.

test/ffmpeg-qsv/transcode/transcoder.py
haribommi: removed this check and enabled SWtoHW scaling.

test/ffmpeg-qsv/transcode/transcoder.py
haribommi: removed this and added HW, SW scaling

test/ffmpeg-qsv/transcode/transcoder.py
haribommi: this code removed as part other review comment to generate yuv directly from output file

test/ffmpeg-qsv/transcode/transcoder.py
haribommi: added code to purge encoded file also

test/ffmpeg-qsv/transcode/transcoder.py
haribommi: implemented decode+scale yuv generation but seeing some issues with multi codec scaling

test/ffmpeg-vaapi/transcode/transcoder.py
haribommi: added vpp requirement check

test/ffmpeg-vaapi/transcode/transcoder.py
haribommi: removed self from variables

test/ffmpeg-qsv/transcode/transcoder.py
```diff
@@ -75,11 +84,19 @@ def validate_spec(self):
...
       codec = output["codec"]
       mode  = output["mode"]
+      width = output.get("width", 0)
+      height  = output.get("height", 0)
       assert mode in ["sw", "hw"], "Invalid test case specification as output mode type not valid"
       oplats, oreq, _ = self.get_requirements_data("encode", codec, mode)
       platforms &= set(oplats)
       requires.append(oreq)
 
+      if width > 0 and height > 0:
```
uartie: What if user only specify width and not height (or vice versa)?  Furthermore, if such a case, either one should default to source dimension anywhere it is used in the generated transcode pipeline commands.

test/ffmpeg-qsv/transcode/transcoder.py
```diff
@@ -120,15 +137,25 @@ def gen_output_opts(self):
...
       _, _, ffparms = self.get_requirements_data("encode", codec, mode)
       assert ffparms is not None, "failed to find a suitable encoder"
 
       ffcodec, ext = ffparms
 
       for channel in xrange(output.get("channels", 1)):
         if "hw" == mode and "sw" == self.mode:
-          opts += " -vf hwupload=extra_hw_frames=64,format=qsv"
+          if width > 0 and height > 0:
+            opts += " -vf 'hwupload=extra_hw_frames=64,format=qsv,vpp_qsv=w={}:h={}'".format(width, height)
```
uartie: Why hard-coded scaling plugin here and other places?  This is not maintainable code.  Scaling plugin should be obtained from the requirements map.

test/ffmpeg-qsv/transcode/transcoder.py
```diff
@@ -173,14 +220,20 @@ def transcode(self):
...
           "{}_{}_{}.yuv".format(self.case, n, channel))
-        call(
-          "ffmpeg -i {} -pix_fmt yuv420p -vframes {}"
-          " -y {}".format(encoded, self.frames, yuv))
+        if width > 0 and height > 0:
+          self.scale_to_yuv(yuv, encoded, output["codec"], output["mode"])
+        else:
+          call(
+            "ffmpeg -i {} -pix_fmt yuv420p -vframes {}"
+            " -y {}".format(encoded, self.frames, yuv))
```
uartie: Why decode to yuv based on "mode" if only scaling is involved, yet always sw decode when scaling is not involved.

test/ffmpeg-qsv/transcode/transcoder.py
```diff
@@ -120,15 +137,25 @@ def gen_output_opts(self):
...
+          if width > 0 and height > 0:
+            opts += " -vf 'hwupload=extra_hw_frames=64,format=qsv,vpp_qsv=w={}:h={}'".format(width, height)
+          else:
+            opts += " -vf hwupload=extra_hw_frames=64,format=qsv"
+        elif "sw" == mode and "hw" == self.mode:
+          if width > 0 and height > 0:
+            opts += " -vf 'hwdownload,format=nv12,scale=w={}:h={}'".format(width, height)
+        else:
+          if width > 0 and height > 0:
+            opts += " -vf 'hwupload=extra_hw_frames=64,format=qsv,vpp_qsv=w={}:h={}'".format(width, height)
```
uartie: There are several common parts to each mode combination and can be de-coupled from scaling.  Seems to me that this can be handled better without the chaos of nested if/else conditions.

test/ffmpeg-qsv/transcode/transcoder.py
```diff
@@ -154,6 +181,26 @@ def check_output(self):
       "hwaccel initialisation returned error", self.output, re.MULTILINE)
     assert m is None, "Failed to use hardware decode"
 
+  # convert scaled test stream to yuv with source width and height
+  def scale_to_yuv(self, yuv, encoded, codec, mode):
```
uartie: I'm not convinced that a dedicated function is necessary.  There are so many common pipeline pieces that are coded over and over in this class.  Much can be captured via code reuse.  Break it down to decode parts, filter parts, and output parts.

uartie:
Maybe https://trac.ffmpeg.org/wiki/Hardware/VAAPI and https://trac.ffmpeg.org/wiki/Hardware/QuickSync examples will help?

test/ffmpeg-qsv/transcode/transcoder.py
haribommi: instead of falling back to source dimensions, shall I assert as width or height invalid value?

uartie:
superseded

PR #60, opened 2019-05-02
Transcode test cleanup, improvements and optimizations

uartie:


test/ffmpeg-vaapi/transcode/transcoder.py
```diff
@@ -17,23 +17,23 @@ class TranscoderTest(slash.Test):
     decode = {
       "avc" : dict(
         sw = (ALL_PLATFORMS, have_ffmpeg_decoder("h264"), "h264"),
-        hw = (AVC_DECODE_PLATFORMS, (True, True), None),
+        hw = (AVC_DECODE_PLATFORMS, have_ffmpeg_decoder("h264"), "h264"),
```
haribommi: is it ok to use SW decoder under HW codec assignment?

uartie: The vaapi hw decoders are built-in ( I've repeated this fact time and time again :) ) ...

This means the `h264` decoder in ffmpeg-vaapi includes both sw decode and hw(vaapi) decode.  The mode that get's used isn't determined until runtime (e.g. --hwaccel vaapi enables the hw path)

PR #66, opened 2019-05-12
ffmpeg-vaapi/vpp: add transpose support

zhuqingliang:

### DIFF
--- a/test/ffmpeg-vaapi/util.py
+++ b/test/ffmpeg-vaapi/util.py
@@ -91,6 +91,18 @@ def map_deinterlace_method(method):
   }.get(method, None)
 
 @memoize
+def map_transpose_direction(degrees, method):
+  return {
+    (270,   "vertical")   : "cclock_flip",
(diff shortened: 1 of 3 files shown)


test/ffmpeg-vaapi/util.py
```diff
@@ -90,6 +90,18 @@ def map_deinterlace_method(method):
...
+  return {
+    (270,   "vertical")   : "cclock_flip",
+    ( 90,            0)   : "clock",
+    (270,            0)   : "cclock",
+    ( 90,   "vertical")   : "clock_flip",
+    (180,            0)   : "reversal",
+    (  0, "horizontal")   : "hflip",
+    (  0,   "vertical")   : "vflip",
+  }.get( (degrees, method), None)
+
```
uartie: I would prefer `None` instead of `0` as user config value for disabling mirror method.

zhuqingliang: updated,  please review.

PR #69, opened 2019-05-14
R2R: ffmpeg-qsv decode tests

haribommi:


test/ffmpeg-qsv/decode/decoder.py
```diff
@@ -14,24 +14,44 @@ class DecoderTest(slash.Test):
...
     get_media().test_call_timeout = vars(self).get("call_timeout", 0)
 
-    self.output = call(
-      "ffmpeg -hwaccel qsv -hwaccel_device /dev/dri/renderD128 -v verbose"
-      " -c:v {ffdecoder} -i {source} -vf 'hwdownload,format={hwformat}'"
-      " -pix_fmt {mformat} -f rawvideo -vsync passthrough"
-      " -vframes {frames} -y {decoded}".format(**vars(self)))
+    if vars(self).get("r2r", None) is not None:
+      assert type(self.r2r) is int and self.r2r > 1, "invalid r2r value"
+      for self.i in xrange(self.r2r):
```
uartie: `self.i` should be changed to `i` since `i` is not needed outside of local/loop scope

haribommi: Agreed, self.i i not global but added i to self to maintain same syntax for 
self.decoded = get_media()._test_artifact("{case}_{width}x{height}_{format}_{i}.yuv".format(**vars(self)))  otherwise it has to be changed to {}_{}x{}_{}_{}.yuv".format(self.case, self.width, self.height, self.format, i)

haribommi: thanks, changed self.i to i

uartie:
Also, please consolidate code similar to #72

haribommi:
> Also, please consolidate code similar to #72

thanks, modified code similar to #72

PR #74, opened 2019-05-17
ffmpeg-vaapi VPP R2R test script changes

haribommi:


haribommi:
@uartie , @FocusLuo and @wangzj0601 - could you please review these changes

test/ffmpeg-vaapi/vpp/brightness.py
```diff
@@ -52,3 +60,29 @@ def compare(k, ref, actual):
...
+@platform_tags(VPP_PLATFORMS)
+def test_r2r(case, level):
+  params = init(spec_r2r, case, level)
+  params.setdefault("r2r", 5)
+  assert type(params["r2r"]) is int and params["r2r"] > 1, "invalid r2r value"
+
+  for i in xrange(params["r2r"]):
+    params["ofile"] = get_media()._test_artifact(
+      "{}_brightness_{}_{}_{}x{}_{}"
+      ".yuv".format(case, params["level"], params["format"], params["width"], params["height"], i))
```
uartie: The following is preferred:
```
params["ofile"] = get_media()._test_artifact(
  "{case}_brightness_{level}_{format}_{width}x{height}_{i}"
  ".yuv".format(case = case, i = i, **params))
```

test/ffmpeg-vaapi/vpp/brightness.py
```diff
@@ -52,3 +60,29 @@ def compare(k, ref, actual):
     get_media().baseline.check_result(
       compare = compare, context = params.get("refctx", []), psnr = psnr)
 
+@slash.requires(have_ffmpeg)
+@slash.requires(have_ffmpeg_vaapi_accel)
+@slash.requires(*have_ffmpeg_filter("procamp_vaapi"))
+@slash.parametrize(*gen_vpp_brightness_parameters(spec_r2r))
+@platform_tags(VPP_PLATFORMS)
+def test_r2r(case, level):
+  params = init(spec_r2r, case, level)
```
uartie: Similar to enc/dec r2r, the first `ofile` setup and `call_ffmpeg` can be the same/shared code with both `test_default` and `test_r2r` tests.  If it helps, you can think of the `md5 ref calculation + for loop` as the check metric part of the r2r test.

haribommi: moved to common code one function to avoid duplicate code

test/ffmpeg-vaapi/vpp/brightness.py
haribommi: thanks, modified it

test/ffmpeg-vaapi/vpp/contrast.py
```diff
@@ -8,33 +8,50 @@
...
     " -pix_fmt {mformat} -vframes {frames} -y {ofile}".format(**params))
 
+def gen_output(case, params):
+  params["ofile"] = get_media()._test_artifact(
+    "{case}_contrast_{level}_{format}_{width}x{height}".format(case = case, **params))
+
+  if params.get("r2r", None) is not None:
+     params["ofile"] =  params["ofile"] + "_r2r"
+
+  params["ofile"] = params["ofile"] + ".yuv"
```
uartie: Do not change the filename that you registered with `get_media()._test_artifact()`.  Build the filename first, then register it with  `get_media()._test_artifact()` last.

uartie: For `get_media()._test_artifact()`:  Basically, you're telling the media plugin that the test is going to produce an artifact file named, say, `abc`.  When the test finishes, the media plugin will automatically remove/save `abc` depending on the artifact-retention setting.  But, if the test actually creates an artifact called `abc_xyz.yuv`, then you've bypassed the media plugin artifact retention mechanism altogether and the file stays around.

haribommi: ok thanks, got the context of it, will correct it

haribommi: corrected the filename with get test artifact

test/ffmpeg-vaapi/vpp/deinterlace.py
```diff
@@ -8,35 +8,79 @@
 from ..util import *
 
 spec = load_test_spec("vpp", "deinterlace")
+spec_r2r = load_test_spec("vpp", "deinterlace", "r2r")
```
wangzj0601: it use md5 for deinterlace test standard in our script, may be we do not need enable it at now

test/ffmpeg-vaapi/vpp/brightness.py
```diff
@@ -8,40 +8,58 @@
...
     " -f rawvideo -pix_fmt {mformat} -s:v {width}x{height} -i {source}"
     " -vf 'format=nv12,hwupload,procamp_vaapi=b={mlevel}"
     ",hwdownload,format=nv12'"
     " -pix_fmt {mformat} -vframes {frames} -y {ofile}".format(**params))
 
+def gen_output(case, params):
+  name = "{case}_brightness_{level}_{format}_{width}x{height}".format(case = case, **params)
+
+  if params.get("r2r", None) is not None:
+     name =  name + "_r2r"
```
uartie: please fix indentation (there's an extra space here and in other gen_output's)

uartie: Also, how about `name += ...` instead?

test/ffmpeg-vaapi/vpp/contrast.py
```diff
@@ -8,33 +8,50 @@
...
     " -f rawvideo -pix_fmt {mformat} -s:v {width}x{height} -i {source}"
     " -vf 'format=nv12,hwupload,procamp_vaapi=c={mlevel}"
     ",hwdownload,format=nv12'"
     " -pix_fmt {mformat} -vframes {frames} -y {ofile}".format(**params))
 
+def gen_output(case, params):
+  name = "{case}_contrast_{level}_{format}_{width}x{height}".format(case = case, **params)
+
+  if params.get("r2r", None) is not None:
+     name =  name + "_r2r"
```
uartie: too many spaces after `=`

test/ffmpeg-vaapi/vpp/brightness.py
```diff
@@ -8,40 +8,58 @@
...
+
+  name = name + ".yuv"
+  params["ofile"] = get_media()._test_artifact("{}".format(name))
```
uartie: `"{}".format(name)` is useless here.

test/ffmpeg-vaapi/vpp/deinterlace.py
haribommi: @uartie , is it ok to keep R2R tests for deinterlace or remove it?

test/ffmpeg-vaapi/vpp/brightness.py
haribommi: thanks, fixed extra space issue and used name +=

test/ffmpeg-vaapi/vpp/contrast.py
haribommi: fixed it

test/ffmpeg-vaapi/vpp/brightness.py
haribommi: removed format and just used name

test/ffmpeg-vaapi/vpp/deinterlace.py
uartie: I don't see anything wrong with keeping it

uartie: @wangzj0601 can you elaborate why you think r2r is not useful now?

wangzj0601: Because we use md5 as standrad for deinterlace test, this can ensure that the md5 must be same every time, I think it is not necessary to enable it at now.

uartie: MD5 is only a default metric in the test_default deinterlace test (i.e. non-r2r).  User configuration can change it to something else if they want.  Thus an R2R test may still be useful to users in some cases.  If we are using MD5 metric for internal config, then it is probably redundant to define any cases for R2R.  Since the test is already written, I don't see any harm in keeping it since user config does not have to enable it.

PR #82, opened 2019-05-23
Fix pixel format support for odd dimensions

uartie:
The frame readers and frame size functions did not properly account for odd dimension in width or height.  This fixes them and also adds some self unit tests to verify correctness of these format algorithms.


wangzj0601:
unit test all passed, looks good to me, thanks

PR #85, opened 2019-05-28
csc: modify the min ssim standard

wangzj0601:
use 1.0 as default min ssim
  use 0.99 as min ssim if reference file exists

### DIFF
--- a/test/ffmpeg-qsv/vpp/csc.py
+++ b/test/ffmpeg-qsv/vpp/csc.py
@@ -40,7 +40,8 @@ def test_default(case, csc):
 
   check_metric(
     # if user specified metric, then use it.  Otherwise, use ssim metric with perfect score
-    metric = params.get("metric", dict(type = "ssim", miny = 1.0, minu = 1.0, minv = 1.0)),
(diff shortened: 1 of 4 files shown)


test/ffmpeg-qsv/vpp/csc.py
```diff
@@ -40,7 +40,8 @@ def test_default(case, csc):
 
   check_metric(
     # if user specified metric, then use it.  Otherwise, use ssim metric with perfect score
-    metric = params.get("metric", dict(type = "ssim", miny = 1.0, minu = 1.0, minv = 1.0)),
+    metric = params.get("metric", dict(type = "ssim", miny = 0.99, minu = 0.99, minv = 0.99)
+      if params.get("reference") else dict(type = "ssim", miny = 1.0, minu = 1.0, minv = 1.0)),
```
uartie: I disagree.  If user specifies a reference, we should assume they want a perfect score by default.  Otherwise, user should set the min params explicitly in the config to override the default.

wangzj0601: ok, let's set the min ssim in config case, so close this PR.

PR #89, opened 2019-06-05
gst-vaapi deinterlace refactor and additions

uartie:
Moved reusable code to a class and added
additional deinterlace tests for avc, mpeg2 and
vc1.

Also, gst-vaapi deinterlace produces output as
field rate (i.e. doubles the number of frames).
Thus, ensure check_metric will take this into
account.


uartie:
On second thought, I think it would be better to deprecate raw input deinterlace.  Thus closing and will resubmit full solution for all middleware.

uartie:
superseded by #90

PR #90, opened 2019-06-06
Redesign Deinterlace Tests

uartie:
* In many cases, raw yuv can't be treated as interlaced content for input to middleware.  Thus, support for raw yuv input is removed.  A warning is emitted if raw yuv cases are defined in the user config.
* Added support for avc, mpeg2 and vc1 deinterlacing.
* Refined deinterlace tests to handle frame vs. field rate output based on middleware capabilities.


uartie:
@FocusLuo @lizhong1008 @xhaihao @wangzj0601 any comments?

lizhong1008:
LGTM

PR #97, opened 2019-07-10
baseline: support for custom reference keys

uartie:
Allow user to specify a reference key in the
test case's refctx setting.  For example:

  refctx = ["key:vp9.decode.10bit"]

will lookup the test case reference value with
the following key:

  vp9.decode.10bit:<test_function(params...)>

This feature enables the ability to share a single
reference value (e.g. MD5) across different
middleware that implement the same test function for
each defined case (e.g. decode tests).


uartie:
comments please!

PR #98, opened 2019-07-16
RFC: WIP: lib/platform: add new platform caps model

uartie:
This new platform capabilities model will allow
for more flexibility and finer-grained control
for tests.  It allows for more descriptive
definitions around constraints and capabilities
for the seemingly endless combinations of
supported formats, resolutions, profiles, codecs,
vpp ops, etc.

Tests can use the caps information to determine
if a specific combination of test runtime
parameters "should" be supported or not by the
currently selected driver/platform.  Then the
tests can skip or proceed with a negative-based
or positive-based execution according to this
information.

For example, vpp tests can call:

  platform.get_caps("vpp", "csc")

...if the return value of the above call is
empty, then it would indicate that we should
not expect vpp csc to be supported by the
platform/driver combination.  Otherwise, a map
of expected hardware input and hardware output
supported formats is returned and can be used
to check if the current test parameters are
included in these supported formats or not.

Another example:

  platform.get_caps("decode", "avc")

...to see if avc decode should be supported
and the expected maximimum supported resolution
(and potentially other constraints added later,
like profiles, etc.).

Eventually, this model will eliminate the
@platform_tags mechanism from this test
framework.  The only benefit of the current tag
mechanism is that it filters out tests before
test execution.  The downside of this "tagging"
is that test runtime parameters aren't known
at the time the tags are evaluated.  This makes
it nearly impossible or hard to filter tests
before execution begins based on the test runtime
parameters.  And therefore, results in various
unsightly hacks in both the test framework and
the user config/spec... for example, see:
  ffmpeg-vaapi/decode/10bit/hevc.py
  gst-vaapi/decode/hevc.py


lib/caps/ICL/iHD
```diff
@@ -0,0 +1,77 @@
...
+  decode  = dict(
+    jpeg    = dict(maxres = (8192, 8192)),
+    mpeg2   = dict(maxres = (4096, 4096)),
+    vc1     = dict(maxres = (4096, 4096)),
+    avc     = dict(maxres = (4096, 4096)),
+    hevc_8  = dict(maxres = (8192, 8192)),
+    hevc_10 = dict(maxres = (4096, 4096)),
+    vp8     = dict(maxres = (4096, 4096)),
+    vp9_8   = dict(maxres = (4096, 4096)),
+    vp9_10  = dict(maxres = (4096, 4096)),
```
xhaihao: AFAIK, maxres should be (8192,8192) for vp9 too on ICL

lib/caps/ICL/iHD
```diff
@@ -0,0 +1,77 @@
...
+  ),
+  encode  = dict(
+    jpeg    = dict(maxres = (8192, 8192)),
+    mpeg2   = dict(maxres = (4096, 4096)),
+    avc     = dict(maxres = (4096, 4096)),
+    hevc_8  = dict(maxres = (8192, 8192)),
+    hevc_10 = dict(maxres = (4096, 4096)),
+    vp8     = dict(maxres = (4096, 4096)),
+    vp9_8   = dict(maxres = (4096, 4096)),
+    vp9_10  = dict(maxres = (4096, 4096)),
```
xhaihao: vp9 encoding is not supported in normal mode on ICL

lib/caps/ICL/iHD
```diff
@@ -0,0 +1,77 @@
...
+  ),
+  vdenc   = dict(
+    avc     = dict(maxres = (4096, 4096)),
+    hevc_8  = dict(maxres = (8192, 8192)),
+    hevc_10 = dict(maxres = (4096, 4096)),
```
xhaihao: vp9 encoding is supported in this mode.

xhaihao:
Due to the limitation of MSDK, YV12 is not supported as output for VPP case, so besides driver/platform combination, we should take the software stack into account.

lib/caps/ICL/iHD
uartie: Yes, good catch.  I think I read the media driver feature E vs. Es backwards :)... will fix.

lib/caps/ICL/iHD
uartie: will fix.

lib/caps/ICL/iHD
uartie: For both 8 and 10 bit? Decode and vdenc?

uartie:
> Due to the limitation of MSDK, YV12 is not supported as output for VPP case, so besides driver/platform combination, we should take the software stack into account.

Agree.  This is already taken into account when we map formats into middleware equivalents at the middleware test implemenation level... and if a mapping is not defined for the middleware level, the case is skipped.  That is, we will map user config format to driver/platform caps then to middleware caps for total support detection.

lib/caps/ICL/iHD
xhaihao: hevc and vp9 use the same HCP unit, so the maximum resolution should be same for hevc and vp9. I am not sure whether 10bit and vdenc support 8K and will double check the media driver.

uartie: Aha, just found this https://github.com/intel/media-driver/blob/master/docs/media_features.md which should answer the question 😃

lib/caps/ICL/iHD
```diff
@@ -0,0 +1,78 @@
...
+### Copyright (C) 2019 Intel Corporation
+###
+### SPDX-License-Identifier: BSD-3-Clause
+###
+
+###
+### kate: syntax python;
+###
+
+# https://github.com/intel/media-driver/blob/master/docs/media_features.md
```
xuguangxin: we have many duplicates here. is it possible to reuse configurations from the previous Gen? like
```
import  GLK/iHD
caps.decode.jjpeg.maxres=res16k
```
If someone does need full details, we can add some command line to print full configuration.

uartie: I think I'd prefer not to inherit any settings here.  I'd like to be able to see all the explicit capabilities of a driver/plat by viewing a single file without needing to follow the import trail or run the code with an option.

If we inherit caps from other caps, then all caps would need to be loaded at runtime and require explicit copying of inherited caps before modifying current caps.  Also, we would need to convert these directories/files into Python modules which I prefer we don't.

test/gst-vaapi/decode/hevc.py
```diff
@@ -14,50 +14,16 @@ class default(DecoderTest):
...
-    vars(self).update(
-      case        = case,
-      gstdecoder  = "h265parse ! vaapih265dec",
-    )
-    self.decode()
-
-  @platform_tags(HEVC_DECODE_8BIT_444_PLATFORMS)
-  @slash.requires(*have_gst_element("vaapih265dec"))
-  @slash.parametrize(("case"), sorted([k for k,v in spec.items() if v["format"] in mapsubsampling("FORMATS_444")]))
-  def test_444(self, case):
```
uartie: TODO: add translation for old baseline test name to new test name

uartie:
All, I have mostly finished converting gst-vaapi tests to use the new caps model.  Please test and provide review.  If it looks good, I can start converting the other middleware tests.

lib/caps/ICL/iHD
xuguangxin: ok, thanks for explaining.

test/gst-vaapi/decode/hevc.py
zhuqingliang: Delete test_* may cause little difficult to validate only 422/444/8k cases and to count 422/444/8k pass rate in hevc decode logs folder.

uartie: For now, you can add a prefix to your test case names (e.g. 444_<case name>) in the user config.  Then use the command-line filter option to filter cases with this prefix.  Later, I plan to add a "category" feature to allow us to assign each case a category in the user config.  Categories will improve our reports and let us group together test cases with common attributes.

uartie:
Work is complete.  Any final reviews before I merge?

test/ffmpeg-vaapi/decode/decoder.py
```diff
@@ -27,11 +27,25 @@ def gen_name(self):
       name += "_r2r"
     return name
 
-  def decode(self):
-    self.mformat = mapformat(self.format)
+  def validate_caps(self):
+    if match_best_format(self.format, self.caps["fmts"]) is None:
```
zhuqingliang: test with " ./full run test/ffmpeg-vaapi/decode" , error happen because some decode caps ( avc /vc1) don't have "fmts" keyvalues , please update.

uartie: What is the error you see?  What is the full command line you used?  The `--platform` command line option is required, so not sure how you got it to run with the command you provided me.

uartie: ah, yes I see I forgot to finish adding "fmts" key to some of the "libs/caps/" files.  I will update.

uartie: Fixed.  Please try again.

PR #100, opened 2019-07-22
baseline: add type judgement for keys of baseline

wangzj0601:
if the keys is not string, such as "tuple", then
  it can not use method "startswith".

### DIFF
--- a/.slashrc
+++ b/.slashrc
@@ -128,7 +128,7 @@ class MediaPlugin(slash.plugins.PluginInterface):
 
   def _get_ref_addr(self, context):
     path, case = slash.context.test.__slash__.address.split(':')
-    keyctx = filter(lambda c: c.startswith("key:"), context)
+    keyctx = filter(lambda c: isinstance(c,str) and c.startswith("key:"), context)
     if len(keyctx):
       key = keyctx[0].lstrip("key:")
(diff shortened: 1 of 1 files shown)


uartie:
In what case does a context include a non-string type?

uartie:
> In what case does a context include a non-string type?

Ah yes, transcode tests.

.slashrc
```diff
@@ -128,7 +128,7 @@ class MediaPlugin(slash.plugins.PluginInterface):
 
   def _get_ref_addr(self, context):
     path, case = slash.context.test.__slash__.address.split(':')
-    keyctx = filter(lambda c: c.startswith("key:"), context)
+    keyctx = filter(lambda c: isinstance(c,str) and c.startswith("key:"), context)
```
uartie: How about just `str(c).startswith("key:")`?

uartie:
Ugh, I guess I can't add commit to PR and squash from GitHub.

Need this fix ASAP so full test runs don't break today.  I merged a slightly modified commit as  b6902e37e4d0ee3768150ba5ebb68d83d727ee45.

Thanks for spotting this issue quickly.

wangzj0601:
Ok, thanks

PR #101, opened 2019-07-25
vp9 8bit 8k decode: add test_highres to handle 8k case

zhuqingliang:
width or height > 4096 will be considered as 8k clips,
and will decode in test_highres.


uartie:
#98 will eliminate the need for this patch.  #98 is still WIP (gst-vaapi has been converted, others are in progress).  Can you help test #98 and provide your thoughts about the change?

PR #103, opened 2019-08-09
test: remove method "weave" for gst-msdk.

wang-zx:
according to iHD driver, it donesn't support method "weave",
 so remove it.


uartie:
See #98

wang-zx:
> See #98

Will the feature available for gst-msdk? I saw the gst-msdk files were not changed.

PR #105, opened 2019-08-15
refactor ffmpeg vpp for code reuse

uartie:
Use common vpp classes for ffmpeg vpp tests


uartie:
@wang-zx please help review

PR #106, opened 2019-08-16
add param for gst-vaapi cmdline

wang-zx:
after add dither=0 for gst-vaapi decode cmdline,
  all hevc-10bitd 444 and 422 cases passed,
  no bad effects on other cases.


uartie:
Should we disable dither in every usage of videoconvert in other tests?

uartie:
@wang-zx any response to my last comment?

wang-zx:
> Should we disable dither in every usage of videoconvert in other tests?

I just ran the gst-vaapi decode tests, the pass rate increased significantly for 10bit. Tests outside gst-vaapi decode may shoudn't change the param.

PR #108, opened 2019-08-23
remove `profile` for gst-msdk mpeg2enc

wang-zx:
Gst-msdk mpeg2enc will automatically determine output profile according to
  mpeg2 spec. If the real profile is not equal to the assignment, it will
  cause error.

### DIFF
--- a/test/gst-msdk/encode/encoder.py
+++ b/test/gst-msdk/encode/encoder.py
@@ -64,8 +64,11 @@ class EncoderTest(slash.Test):
 #    if vars(self).get("looplvl", None) is not None:
 #      opts += " loop-filter-level={looplvl}"
 
-    if self.codec not in ["jpeg", "vp8", ]:
+    if self.codec not in ["jpeg", "vp8", "mpeg2"]:
       opts += " ! {gstmediatype},profile={mprofile}"
+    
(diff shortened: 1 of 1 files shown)


uartie:
Since msdkmpeg2enc exposes this setting to the user with multiple choices, we should be allowed to test it.  If user setting causes errors, then that is a bug in gst-msdk.  This patch just hides the bug.

uartie:
In other schools-of-thought, an error might be deemed as the appropriate action here.  We'll want to clarify that with the community of that project.  In such a case, we should add support to the tests to allow negative outcomes to be treated as "pass" (e.g. allow user-configurable expectations).  The other way around it is to avoid defining a test case for it in the user config.

fuweitax:
> Since msdkmpeg2enc exposes this setting to the user with multiple choices, we should be allowed to test it. If user setting causes errors, then that is a bug in gst-msdk. This patch just hides the bug.

Hi, Artie. In fact, the actual mpeg2 profile is determined by MediaSDK based on the input parameters (such as width, height or b-frames), but not the "profile=xxx" specified on the command line. However, if the actual profile is inconsistent with the specified profile, this will cause gstreamer errors, so we may need to remove the "profile" from the command line.

PR #112, opened 2019-09-05
vpp/denoise: add Luma into ICL denoise cap

zhuqingliang:
On ICL, for both gstreamer and ffmpeg,
  denoise will only influence Luma,
  and for the other platforms Luma and Chroma both changed

### DIFF
--- a/lib/caps/ICL/iHD
+++ b/lib/caps/ICL/iHD
@@ -64,6 +64,7 @@ caps = dict(
     denoise     = dict(
       ifmts = ["NV12", "P010", "YUY2"],
       ofmts = ["NV12", "P010", "YUY2"],
+      chroma = False,
     ),
     scale       = dict(
       ifmts = ["NV12", "YV12", "I420", "P010", "YUY2", "UYVY", "Y210", "AYUV", "Y410", "BGRA"],
--- a/test/ffmpeg-qsv/vpp/denoise.py
+++ b/test/ffmpeg-qsv/vpp/denoise.py
@@ -39,8 +39,12 @@ class default(VppTest):
     def compare(k, ref, actual):
       assert ref is not None, "Invalid reference value"
       assert abs(ref[-3] - actual[-3]) < 0.2, "Luma (Y) out of baseline range"
-      assert abs(ref[-2] - actual[-2]) < 0.2, "Cb (U) out of baseline range"
-      assert abs(ref[-1] - actual[-1]) < 0.2, "Cr (V) out of baseline range"
+      if self.caps.get("chroma", True):
+        assert abs(ref[-2] - actual[-2]) < 0.2, "Cb (U) out of baseline range"
+        assert abs(ref[-1] - actual[-1]) < 0.2, "Cr (V) out of baseline range"
+      else:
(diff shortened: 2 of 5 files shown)


test/ffmpeg-qsv/vpp/denoise.py
```diff
@@ -38,9 +38,12 @@ def check_metrics(self):
 
     def compare(k, ref, actual):
       assert ref is not None, "Invalid reference value"
-      assert abs(ref[-3] - actual[-3]) < 0.2, "Luma (Y) out of baseline range"
-      assert abs(ref[-2] - actual[-2]) < 0.2, "Cb (U) out of baseline range"
-      assert abs(ref[-1] - actual[-1]) < 0.2, "Cr (V) out of baseline range"
+      effects = self.caps.get("effect", [ "Luma", "Chroma"])
+      if "Luma" in effects:
```
uartie: Luma should always be supported until there is evidence claiming otherwise.  So let's always check Luma unconditionally.

lib/caps/ICL/iHD
```diff
@@ -64,6 +64,7 @@ caps = dict(
     denoise     = dict(
       ifmts = ["NV12", "P010", "YUY2"],
       ofmts = ["NV12", "P010", "YUY2"],
+      effect = ["Luma"],
```
uartie: Can we use `chroma = False` instead?  Then use `self.caps.get("chroma", True)` to check for the cap.  This would follow the other style better (e.g. `multislice` in vdenc avc lib/caps/KBL/i965).

test/ffmpeg-qsv/vpp/denoise.py
zhuqingliang: Yes , you are right.

zhuqingliang:
updated . Please review. Thanks.

test/ffmpeg-qsv/vpp/denoise.py
```diff
@@ -39,8 +39,9 @@ def check_metrics(self):
     def compare(k, ref, actual):
       assert ref is not None, "Invalid reference value"
       assert abs(ref[-3] - actual[-3]) < 0.2, "Luma (Y) out of baseline range"
-      assert abs(ref[-2] - actual[-2]) < 0.2, "Cb (U) out of baseline range"
-      assert abs(ref[-1] - actual[-1]) < 0.2, "Cr (V) out of baseline range"
+      if self.caps.get("chroma", True):
+        assert abs(ref[-2] - actual[-2]) < 0.2, "Cb (U) out of baseline range"
+        assert abs(ref[-1] - actual[-1]) < 0.2, "Cr (V) out of baseline range"
```
uartie: If chroma is False, then we should also check that it did not change.  For example:
```python

if self.caps.get("chroma", True):
  ...
else:
  assert actual[-2] == 100, "Cb(U) changed, but caps don't support DENOISE chroma"
  assert actual[-1] == 100, "Cr(V) changed, but caps don't support DENOISE chroma"
```
This will let us catch if driver adds support for it in the future or introduces a bug that changes chroma unintentionally.

zhuqingliang:
updated. Please review . Thanks.

PR #118, opened 2019-09-20
gst-msdk/transcode: eliminate videoconvert and use NV12 default

uartie:
The videoconvert element can slow down transcode pipeline by
~6x in many cases.  Thus, eliminate videoconvert usage
where possible and use NV12 as default comparison format.


uartie:
This change can also fix some of our internal transcode test failures.

PR #124, opened 2019-10-16
gst-msdk/vpp: add crop test

uartie:
Depends on https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/issues/1097

### DIFF
--- a/test/gst-msdk/vpp/crop.py
+++ b/test/gst-msdk/vpp/crop.py
@@ -1,0 +1,46 @@
+###
+### Copyright (C) 2019 Intel Corporation
+###
+### SPDX-License-Identifier: BSD-3-Clause
+###
+
+from ....lib import *
+from ..util import *
+from .vpp import VppTest
+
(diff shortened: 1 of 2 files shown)


uartie:
> This patch supports python2 or python3 ?

vaapi-fits has not been ported to python 3, yet.  That is your task.

uartie:
> > This patch supports python2 or python3 ?
> 
> vaapi-fits has not been ported to python 3, yet. That is your task.

either way, this patch should work with both versions of python

zhoujd:
well, all the patches with python3 and python2  compatibility will be good .

zhoujd:
It looks good to me. quick checked on local CI, will apply the patch if no objection

uartie:
> It looks good to me. quick checked on local CI, will apply the patch if no objection

Please don't merge until gst-plugins-bad crop patch is merged (see "Depends" link above).

PR #128, opened 2019-10-25
avc/hevc: add max min frame qp test for cbr/vbr encode

zhuqingliang:

### DIFF
--- a/test/ffmpeg-qsv/encode/encoder.py
+++ b/test/ffmpeg-qsv/encode/encoder.py
@@ -55,6 +55,10 @@ class EncoderTest(slash.Test):
     if vars(self).get("ladepth", None) is not None:
       opts += " -look_ahead 1"
       opts += " -look_ahead_depth {ladepth}"
+    if vars(self).get("qmax", None) is not None:
+      opts += " -qmax {qmax}"
+    if vars(self).get("qmin", None) is not None:
+      opts += " -qmin {qmin}"
 
     opts += " -vframes {frames} -y {encoded}"
 
@@ -88,6 +92,10 @@ class EncoderTest(slash.Test):
       name += "-{ladepth}"
     if vars(self).get("r2r", None) is not None:
       name += "-r2r"
+    if vars(self).get("qmax", None) is not None:
+      name += "-{qmax}"
+    if vars(self).get("qmin", None) is not None:
+      name += "-{qmin}"
 
     return name
 
--- a/test/ffmpeg-qsv/encode/hevc.py
+++ b/test/ffmpeg-qsv/encode/hevc.py
@@ -89,7 +89,8 @@ class cqp_lp(HEVC8EncoderTest):
     self.encode()
 
 class cbr(HEVC8EncoderTest):
-  def init(self, tspec, case, gop, slices, bframes, bitrate, fps, profile):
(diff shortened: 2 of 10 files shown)


zhoujd:
It looks good to me.

test/ffmpeg-qsv/encode/encoder.py
```diff
@@ -55,6 +55,10 @@ def gen_output_opts(self):
     if vars(self).get("ladepth", None) is not None:
       opts += " -look_ahead 1"
       opts += " -look_ahead_depth {ladepth}"
+    if vars(self).get("qmax", None) is not None:
+      opts += " -qmax {qmax}"
+    if vars(self).get("qmin", None) is not None:
+      opts += " -qmin {qmin}"
```
uartie: qmin and qmax should be handled in gen_name, too.

test/ffmpeg-vaapi/encode/encoder.py
```diff
@@ -65,6 +65,10 @@ def gen_output_opts(self):
     if vars(self).get("level", None) is not None:
       self.level /= 10.0
       opts += " -level {level}"
+    if vars(self).get("qmax", None) is not None:
+      opts += " -qmax {qmax}"
+    if vars(self).get("qmin", None) is not None:
+      opts += " -qmin {qmin}"
```
uartie: qmax and qmin should be handled in gen_name, too.

test/gst-vaapi/encode/encoder.py
```diff
@@ -63,6 +63,10 @@ def gen_output_opts(self):
       opts += " sharpness-level={loopshp}"
     if vars(self).get("looplvl", None) is not None:
       opts += " loop-filter-level={looplvl}"
+    if vars(self).get("qmax", None) is not None:
+      opts += " max-qp={qmax}"
+    if vars(self).get("qmin", None) is not None:
+      opts += " min-qp={qmin}"
```
uartie: same as previous comments.

test/ffmpeg-qsv/encode/hevc.py
```diff
@@ -89,7 +89,8 @@ def test_r2r(self, case, gop, slices, qp, quality, profile):
     self.encode()
 
 class cbr(HEVC8EncoderTest):
-  def init(self, tspec, case, gop, slices, bframes, bitrate, fps, profile):
+  def init(self, tspec, case, gop, slices, bframes, bitrate, fps, profile,
+          qmax = None, qmin = None):
```
uartie: AFAICT, the hevc encoder in ffmpeg-qsv does not consider/handle qmin or qmax.  Please correct me if I'm wrong.

zhuqingliang: Can we add  test feature in vaapi-fits , then to file a bug to developer ? I won't add reference for all test case.

uartie: No.  If middleware does not support the feature, then a test for the feature should not be added.  Please consider the potential that there may be other users of this test framework.  And if we add the test, then those users could generate their own cases and baselines and erroneously assume they are correct.

If you believe the middleware should support this feature, please file a bug in the middleware.

PR #133, opened 2019-10-30
encode: add ICQ mode test

zhuqingliang:
add avc/hevc_8bit/hevc_10bit ICQ mode encode
test for ffmpeg-vaapi and ffmpeg-qsv

### DIFF
--- a/lib/parameters.py
+++ b/lib/parameters.py
@@ -54,6 +54,70 @@ def gen_avc_cqp_parameters(spec, profiles):
   params = gen_avc_cqp_variants(spec, profiles)
   return keys, params
 
+def gen_avc_icq_variants(spec, profiles):
+  for case, params in spec.iteritems():
+    for variant in copy.deepcopy(params.get("icq", [])):
+      uprofile = variant.get("profile", None)
+      cprofiles = [uprofile] if uprofile else profiles
+      for profile in cprofiles:
+        yield [
+          case, variant["gop"], variant["slices"], variant["bframes"],
+          variant["quality"], variant.get("icq_quality"), profile
+        ]
+
+def gen_avc_icq_parameters( spec, profiles):
+  keys = ("case", "gop", "slices", "bframes", "quality", "icq_quality", "profile")
+  params = gen_avc_icq_variants(spec, profiles)
+  return keys, params
+
+def gen_avc_icq_la_variants(spec, profiles):
+  for case, params in spec.iteritems():
+    for variant in copy.deepcopy(params.get("icq_la", [])):
+      uprofile = variant.get("profile", None)
+      cprofiles = [uprofile] if uprofile else profiles
+      for profile in cprofiles:
+        yield [
+          case, variant["gop"], variant["slices"], variant["bframes"],
+          variant["quality"], variant.get("icq_quality"), profile, variant.get("ladepth")
+        ]
+
+def gen_avc_icq_la_parameters( spec, profiles):
+  keys = ("case", "gop", "slices", "bframes", "quality", "icq_quality", "profile", "ladepth")
+  params = gen_avc_icq_la_variants(spec, profiles)
+  return keys, params
+
+def gen_avc_icq_lp_variants(spec, profiles):
+  for case, params in spec.iteritems():
+    for variant in copy.deepcopy(params.get("icq_lp", [])):
+      uprofile = variant.get("profile", None)
+      cprofiles = [uprofile] if uprofile else profiles
+      for profile in cprofiles:
+        yield [
+          case, variant["gop"], variant["slices"], variant["bframes"],
+          variant["quality"], variant.get("icq_quality"), profile
+        ]
+
+def gen_avc_icq_lp_parameters( spec, profiles):
+  keys = ("case", "gop", "slices", "bframes", "quality", "icq_quality", "profile")
+  params = gen_avc_icq_lp_variants(spec, profiles)
+  return keys, params
+
+def gen_avc_icq_lpla_variants(spec, profiles):
+  for case, params in spec.iteritems():
+    for variant in copy.deepcopy(params.get("icq_lpla", [])):
+      uprofile = variant.get("profile", None)
+      cprofiles = [uprofile] if uprofile else profiles
+      for profile in cprofiles:
+        yield [
+          case, variant["gop"], variant["slices"], variant["bframes"],
+          variant["quality"], variant.get("icq_quality"), profile, variant.get("ladepth")
+        ]
+
+def gen_avc_icq_lpla_parameters( spec, profiles):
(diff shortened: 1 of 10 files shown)


test/ffmpeg-qsv/encode/encoder.py
```diff
@@ -36,10 +36,7 @@ def gen_output_opts(self):
       else:
         opts += " -q {qp}"
     if vars(self).get("quality", None) is not None:
-      if self.codec in ["jpeg",]:
-        opts += " -global_quality {quality}"
-      else:
-        opts += " -preset {quality}"
+      opts += " -global_quality {quality}"
```
uartie: The `preset` and `global_quality` options have different meanings in ffmpeg-qsv.  This will change the behavior of existing tests.  In the context of vaapi-fits, we interpret the"quality" parameter in the encode tests as the "preset" (i.e. compression level 1 thru 7) for all middleware encoders (except for jpeg codec). 

The problem comes from the fact that every middleware uses a different name for setting up the "compression level".  FFmpeg-qsv uses `preset`, ffmpeg-vaapi uses `quality`, gst-msdk uses `target-usage` and gst-vaapi uses `quality-level` to set the "compression level".  So, at vaapi-fits level, it was necessary to normalize the the name to a single meaning (i.e. "quality").  Perhaps "tu" (i.e. target-usage) would have been a better normalized name choice to avoid confusion.  Maybe it's time to change that!  Changing it is cumbersome, though, since it would require writing a test name translation for the baselines and updating all user test configs.  The other problem is we don't have control over other peoples test configs and it would mess them up... although, I'm not sure if others are using this test framework, yet.  Ideally, we should add some sort of versioning scheme for test config and baseline files so that we can provide automatic translation support.

 It's a mess and confusing, I know!  But for now, "quality" **must** be interpreted as the compression level (except for jpeg codec) in vaapi-fits across all middleware tests for proper consistency.

test/ffmpeg-qsv/encode/10bit/hevc.py
```diff
@@ -88,6 +88,58 @@ def test_r2r(self, case, gop, slices, qp, quality, profile):
...
+  @slash.parametrize(*gen_hevc_icq_parameters(spec, ['main10']))
+  def test(self, case, gop, slices, bframes, quality, profile):
+    self.init(spec, case, gop, slices, bframes, quality, profile)
+    self.encode()
+
+  @slash.requires(*platform.have_caps("encode", "hevc_10"))
+  @slash.requires(*have_ffmpeg_encoder("hevc_qsv"))
+  @slash.requires(*have_ffmpeg_decoder("hevc_qsv"))
+  @slash.parametrize(*gen_hevc_icq_la_parameters(spec, ['main10']))
+  def test_la(self, case, gop, slices, bframes, quality, profile, ladepth):
```
uartie: The convention has been to define separate classes for non-lowpower vs. lowpower vs. look-ahead.  For example, `class vbr` vs. `class vbr_lp` vs. `class vbr_la`.  Each convention has its pros and cons.  But please don't diverge from the historical convention.

zhuqingliang: Ok, I will follow the convention to change my code.

test/ffmpeg-qsv/encode/encoder.py
zhuqingliang: So, can you show me how to change the code here ?

uartie: Maybe you can introduce a different parameter to influence global_quality setting for ICQ tests.

uartie: FWIW, ICQQuality is the field name in the msdk mfx structure.  In ffmpeg-qsv, `mfx.ICQQuality = avctx->global_quality`.  In gst-msdk, `param->ICQQuality = thiz->qpi`.  AFAICT, valid range is [1, 51].

So how about "icq_quality" as the vaapi-fits normalized param name which can be translated to the appropriate user param at the middleware test level.

zhuqingliang: PR updated, please review.

lib/parameters.py
```diff
@@ -54,6 +54,70 @@ def gen_avc_cqp_parameters(spec, profiles):
...
+      uprofile = variant.get("profile", None)
+      cprofiles = [uprofile] if uprofile else profiles
+      for profile in cprofiles:
+        yield [
+          case, variant["gop"], variant["slices"], variant["bframes"],
+          variant.get("icq_quality"), profile
+        ]
+
+def gen_avc_icq_parameters( spec, profiles):
+  keys = ("case", "gop", "slices", "bframes", "icq_quality", "profile")
```
uartie: Is "quality" (aka. target-usage (TU) or preset) parameter irrelevant for ICQ?  Can't we have both icq_quality and quality?

test/ffmpeg-qsv/encode/encoder.py
```diff
@@ -167,7 +169,10 @@ def check_output(self):
     assert m is not None, "It appears that the QSV plugin did not load"
 
     if vars(self).get("ladepth", None) is not None:
-      m = re.search("Using the VBR with lookahead \(LA\) ratecontrol method", self.output, re.MULTILINE)
+      if "vbr" == self.rcmode:
+        m = re.search("Using the VBR with lookahead \(LA\) ratecontrol method", self.output, re.MULTILINE)
+      elif "icq" == self.rcmode:
+        m = re.search("RateControlMethod: LA_ICQ", self.output, re.MULTILINE)
```
uartie: What about ICQ without LA?

Would it be better to search for rate control messages generated by the qsvenc.c select_rc_mode function instead (https://github.com/intel-media-ci/ffmpeg/blob/b414cff6304071d5ef1e2b0a1b407e4fe3f58092/libavcodec/qsvenc.c#L373)?

test/ffmpeg-vaapi/encode/encoder.py
```diff
@@ -46,6 +46,8 @@ def gen_output_opts(self):
         opts += " -global_quality {quality}"
       else:
         opts += " -quality {quality}"
+    if vars(self).get("icq_quality", None) is not None:
+      opts += " -quality {icq_quality}"
```
uartie: I thought `-quality` in ffmpeg-vaapi is only used to set the target-usage?  How can it influence icq_quality?

AFAICT, you need to set `-global_quality` (https://github.com/intel-media-ci/ffmpeg/blob/b414cff6304071d5ef1e2b0a1b407e4fe3f58092/libavcodec/vaapi_encode.c#L1523).

test/ffmpeg-qsv/encode/encoder.py
zhuqingliang: Yes, of course. PR update, please review , thanks.

PR #137, opened 2019-10-31
encode: add max frame size test

zhuqingliang:
add avc/hevc_8bit max frame size test


lib/parameters.py
```diff
@@ -82,6 +82,22 @@ def gen_avc_cbr_parameters(spec, profiles):
   params = gen_avc_cbr_variants(spec, profiles)
   return keys, params
 
+def gen_avc_cbr_mfs_variants(spec, profiles):
+  for case, params in spec.iteritems():
```
zhoujd: Change spec.iteritems() to iter(spec.items()) for python3

PR #138, opened 2019-11-05
Port from python2 to python3

zhoujd:


lib/metrics.py
```diff
@@ -124,51 +125,53 @@ def __compare_ssim(planes):
...
 def __compare_nrmse(planes):
   a, b = planes
-  return skimage.measure.compare_nrmse(a, b, norm_type = "Euclidean")
+  with expected_warnings(['DEPRECATED']):
+    return skimage.measure.compare_nrmse(a, b, norm_type = "Euclidean")
 
 @timefn("nrmse")
 def calculate_nrmse(filename1, filename2, width, height, nframes = 1, fourcc = "I420"):
-  return RawMetricAggregator(max).calculate(
+  return list(map(float, RawMetricAggregator(max).calculate(
```
uartie: It would be better to have RawMetricAggregator::calculate function return this result type.

lib/metrics.py
```diff
@@ -124,51 +125,53 @@ def __compare_ssim(planes):
   a, b = planes
   if a is None or b is None: # handle Y800 case
     return 1.0
-  return skimage.measure.compare_ssim(a, b, win_size = 3)
+  return skimage.metrics.structural_similarity(a, b, win_size = 3)
```
uartie: Why the function name change?  Did it change in newer version of skimage?  If so, please support both skimage versions during import.  That is, if user has old version installed, then it should still work.  You can assign the appropriate function to a global function pointer variable depending on the imported version, then just use the global variable function pointer here.

.slashrc
```diff
@@ -131,15 +131,15 @@ class MediaPlugin(slash.plugins.PluginInterface):
...
+    keyctx = list(filter(lambda c: str(c).startswith("key:"), context))
     if len(keyctx):
       key = keyctx[0].lstrip("key:")
     else:
       key = os.path.relpath(path, self.mypath)
     return ':'.join([key, case])
 
   def _set_test_details(self, **kwargs):
-    for k, v in kwargs.iteritems():
+    for k, v in iter(kwargs.items()):
```
uartie: Is `iter` actually necessary?  I think `kwargs.items()` is enough.  Same everywhere else.

lib/platform.py
```diff
@@ -25,7 +25,13 @@ def load_caps():
     )
   )
   if os.path.exists(capsfile):
-    execfile(capsfile, namespace)
+    global_namespace = {
+      "__file__": capsfile,
+      "__name__": "__main__",
+    }
+    with open(capsfile, 'rb') as file:
```
uartie: `as f` instead of `as file` please

lib/platform.py
```diff
@@ -25,7 +25,13 @@ def load_caps():
...
+      exec(compile(file.read(), capsfile, 'exec'), global_namespace, namespace)
```
uartie: Hmmm... any reason why `exec(f.read())` wouldn't work here?

uartie:
`./vaapi-fits list` fails

lib/metrics.py
uartie: It looks like several of skimage.measure functions are deprecated since 0.16 (https://scikit-image.org/docs/dev/api_changes.html#version-0-16) .  We should do something like:

```python
try:
  # try skimage >= 0.16, first
  from skimage.metrics import structural_similarity as skimage_ssim
except:
  from skimage.measure import compare_ssim as skimage_ssim
```

...then in `__compare_ssim` we just call `skimage_ssim`.

The same needs to be done for mse, nrmse and psnr metrics, too (https://scikit-image.org/docs/dev/release_notes_and_installation.html#deprecations).

Additionally, this change has nothing to do with py2to3 porting.  Please submit as a separate pull-request.

zhoujd:
will do the change and resend the patches

lib/platform.py
zhoujd: yes, exec(f.read()) works, but compile may make it run faster. will do the change

lib/metrics.py
zhoujd: will do the change

.slashrc
zhoujd: yes, kwargs.items works, will do the change

lib/platform.py
zhoujd: will do change

lib/metrics.py
zhoujd: Yes, changed RawMetricAggregator::calculate return list

zhoujd: Another PR for  this: https://github.com/intel/vaapi-fits/pull/145, please help to review it.

lib/metrics.py
zhoujd: Another PR for this: #145, please help to review it.

config/default
```diff
@@ -10,17 +10,23 @@ import tarfile
...
 for file in os.listdir(os.path.dirname(config)):
   if fnmatch.fnmatch(file, "*.default"):
-    execfile(os.path.join(os.path.dirname(config), file))
+    global_namespace = {
+      "__file__": config,
+      "__name__": "__main__",
+    }
+    local_namespace = locals()
+    with open(os.path.join(os.path.dirname(config), file), 'rb') as f:
+      exec(f.read(), global_namespace, local_namespace)
```
uartie: Can we use the current default globals and locals parameter behavior?

lib/platform.py
```diff
@@ -63,6 +69,8 @@ def have_caps(*args):
 
 def info():
   import platform
+  import distro
```
uartie: `distro` is a non-standard python module.  Thus, please use try...except here.  And if distro is unavailable, try fallback to platform.dist or a lambda that returns "unknown"...

```python
try:
  from distro import linux_distribution as linux_dist
except:
  try:
    from platform import dist as linux_dist
  except:
    linux_dist = lambda: "unknown"
```

.slashrc
```diff
@@ -326,15 +326,22 @@ config = os.environ.get(
...
 assert os.path.exists(config)
-execfile(config)
+
+global_namespace = {
+  "__file__": config,
+  "__name__": "__main__",
+}
+local_namespace = locals()
+with open(config, 'rb') as file:
+  exec(file.read(), global_namespace, local_namespace)
```
uartie: Can we use default globals and locals parameters... that is, `exec(file.read())` should be enough?

lib/metrics.py
```diff
@@ -101,20 +101,18 @@ def __append(self, func, iterable):
   def __get(self):
     self.__collect_async()
     result = list(itertools.chain(*self.results))
-    return map(
-      lambda v: round(v, 4), (
+    return [
```
uartie: We will lose the rounding that we desire.  Please just wrap the previous code with a `list()` cast.

.slashrc
```diff
@@ -326,15 +326,22 @@ config = os.environ.get(
...
   os.path.abspath(os.path.join(media.mypath, "config", "default")))
 assert os.path.exists(config)
-execfile(config)
+
+global_namespace = {
+  "__file__": config,
+  "__name__": "__main__",
+}
+local_namespace = locals()
+with open(config, 'rb') as file:
```
uartie: `as f` please.

lib/platform.py
```diff
@@ -25,7 +25,13 @@ def load_caps():
     )
   )
   if os.path.exists(capsfile):
-    execfile(capsfile, namespace)
+    global_namespace = {
+      "__file__": capsfile,
+      "__name__": "__main__",
+    }
+    with open(capsfile, 'rb') as f:
+      exec(f.read(), global_namespace, namespace)
```
uartie: Just use `exec(f.read())`... I don't think setting globals and locals is necessary.

uartie: Wait, I guess you will need `exec(f.read(), namespace)` instead.

lib/metrics.py
zhoujd: how about use float(round(v, 4)) for v in (

config/default
zhoujd: fixed the warning, and changed for exec

lib/platform.py
zhoujd: did the change

.slashrc
zhoujd: did the change

lib/metrics.py
zhoujd: did the change

lib/platform.py
zhoujd: did the change

.slashrc
zhoujd: did the change

uartie:
Please cleanup your commits...

There are several patches that should be squashed together:

* 43e1d6440424 and 04f221382bf7
* fcff27a6394f and 6d34c9a4e6b7
* bb97daa5a488 and 68a035d9c33d
* 8e965bf7f95d and 9e5368f72adf and 59d746ad89a0

Also, in 59d746ad89a0, please separate the range, import style, and except style changes into their proper commits.  In 68a035d9c33d, please separate the iteritems, except style and string handling into their proper commits.

Additionally, please revisit your commit messages to ensure the titles/descriptions match the actual changes in the commit.

In general, you shouldn't create a separate commit to fix a previous commit in the same series (pull-request)... just squash the fix with the original commit.

uartie:
Please cleanup your commits...

There are several patches that should be squashed together:

* `"lib/framereader.py: fix got float for P210"` and `"lib/framereader.py: replace / with //"`
* `"lib/platform: use try...except for distro"` and `"lib/platform: replace platform.dist() with distro package"`
* `"exec use default namespace"` and `"replace execfile with exec"`
* `"lib/metrics: return list with round(v, 4)"` and `"lib/metrics: change calculate() return list"` and `"lib/metrics: update for py3 calling style"`

Also, in `"lib/metrics: update for py3 calling style"`, please separate the range, import style, and except style changes into their proper commits.  In `"replace execfile with exec"`, please separate the iteritems, except style and string handling into their proper commits.

Additionally, please revisit your commit messages to ensure the titles/descriptions match the actual changes in the commit.

In general, you shouldn't create a separate commit to fix a previous commit in the same series (pull-request)... just squash the fix with the original commit.  And each commit should focus on one thing/topic.

zhoujd:
> * `"exec use default namespace"` and `"replace execfile with exec"`
will change

zhoujd:
> Please cleanup your commits...
> 
> There are several patches that should be squashed together:
> 
> * `"lib/framereader.py: fix got float for P210"` and `"lib/framereader.py: replace / with //"`
> * `"lib/platform: use try...except for distro"` and `"lib/platform: replace platform.dist() with distro package"`
> * `"exec use default namespace"` and `"replace execfile with exec"`
> * `"lib/metrics: return list with round(v, 4)"` and `"lib/metrics: change calculate() return list"` and `"lib/metrics: update for py3 calling style"`
> 
> Also, in `"lib/metrics: update for py3 calling style"`, please separate the range, import style, and except style changes into their proper commits. In `"replace execfile with exec"`, please separate the iteritems, except style and string handling into their proper commits.
> 
> Additionally, please revisit your commit messages to ensure the titles/descriptions match the actual changes in the commit.
> 
> In general, you shouldn't create a separate commit to fix a previous commit in the same series (pull-request)... just squash the fix with the original commit. And each commit should focus on one thing/topic.

will change

zhoujd:
> Please cleanup your commits...
> 
> There are several patches that should be squashed together:
> 
> * `"lib/framereader.py: fix got float for P210"` and `"lib/framereader.py: replace / with //"`
> * `"lib/platform: use try...except for distro"` and `"lib/platform: replace platform.dist() with distro package"`
> * `"exec use default namespace"` and `"replace execfile with exec"`
> * `"lib/metrics: return list with round(v, 4)"` and `"lib/metrics: change calculate() return list"` and `"lib/metrics: update for py3 calling style"`
> 
> Also, in `"lib/metrics: update for py3 calling style"`, please separate the range, import style, and except style changes into their proper commits. In `"replace execfile with exec"`, please separate the iteritems, except style and string handling into their proper commits.
> 
> Additionally, please revisit your commit messages to ensure the titles/descriptions match the actual changes in the commit.
> 
> In general, you shouldn't create a separate commit to fix a previous commit in the same series (pull-request)... just squash the fix with the original commit. And each commit should focus on one thing/topic.

did the change

lib/baseline.py
```diff
@@ -7,7 +7,7 @@
 import json
 import os
 import slash
-from common import get_media
+from .common import get_media
```
uartie: This change is in the wrong commit.  Please combine all "import" syntax changes into a single commit, together.

lib/parameters.py
```diff
@@ -8,7 +8,7 @@
 import itertools
 
 def format_value(value, **params):
-  from common import get_media
+  from .common import get_media
```
uartie: This change is in the wrong commit. Please combine all "import" syntax changes into a single commit, together.

tools/genassets.py
```diff
@@ -1,4 +1,4 @@
-#!/bin/env python2
+#!/usr/bin/env python3
```
uartie: This change is in the wrong commit. Please combine all "/usr/bin/env python3" changes into a single commit, together.

config/default
```diff
@@ -10,17 +10,18 @@ import tarfile
 assets = os.path.join(media.mypath, "assets")
 
 if not os.path.exists(assets):
-  print "Extracting assets..."
+  print("Extracting assets...")
   try:
     with tarfile.open("{}.tbz2".format(assets), "r:*") as tf:
       tf.extractall(media.mypath)
-  except tarfile.ReadError, e:
+  except tarfile.ReadError as e:
```
uartie: Change is in wrong commit.  Please put all "except" fixes into a single commit, together.  The commit should only address "except"... don't combine it with other changes.

lib/common.py
```diff
@@ -119,41 +119,40 @@ def timeout(proc):
...
-
-  if timeout.triggered:
-    error = True
-    get_media()._report_call_timeout()
-    message = "CALL TIMEOUT: timeout after {} seconds (pid: {}).".format(
-      timer.interval, proc.pid)
-  elif proc.returncode != 0:
-    error = True
-    message = "CALL ERROR: failed with exitcode {} (pid: {})".format(proc.returncode, proc.pid)
+  with subprocess.Popen("exec " + command,
```
uartie: This is not backwards compatible with python 2.  This means our internal smoke/full test configs will need to be updated to python3 and merged atomically with this PR.

I'd prefer we keep all python3 changes backwards compatible with python2 for at least a few months... so any external dependents (i.e. smoke/full or other users?) have time to migrate.

So for now, it would be better to keep the original syntax (i.e. don't use `with` on popen).  Instead, to satisfy both py2 and py3, all we need to do is call `proc.stdin.close()` and `proc.stdout.close()` in the `finally:` block after `reader.join(30)`.

zhoujd:
> Commit `"lib/framereader.py: replace / with //"` ([143b65d](https://github.com/intel/vaapi-fits/commit/143b65d680781f15eeca1679941796a4d2f2724a)):
> 
> * please remove `"lib/framereader.py: fix got float for P210"` from the commit message, it makes no sense.
> 
> Please go back and review your commits again. You still have several commits that mix together unrelated changes. Hint: use "git show" on each commit to make sure each commit makes sense (i.e. matches the commit message and doesn't intermix other kinds of changes).

Actually, I have drop `"lib/framereader.py: fix got float for P210"` commit, and rebased code for commit logs, on my local the commits looks good.

Did you use following command lines and check locally.
git fetch origin pull/138/head:pr-138
git checkout pr-138.

Anyway, I will continue split commits.

lib/common.py
zhoujd: will try on local first

uartie:
> > Commit `"lib/framereader.py: replace / with //"` ([143b65d](https://github.com/intel/vaapi-fits/commit/143b65d680781f15eeca1679941796a4d2f2724a)):
> > 
> > * please remove `"lib/framereader.py: fix got float for P210"` from the commit message, it makes no sense.
> > 
> > Please go back and review your commits again. You still have several commits that mix together unrelated changes. Hint: use "git show" on each commit to make sure each commit makes sense (i.e. matches the commit message and doesn't intermix other kinds of changes).
> 
> Actually, I have drop `"lib/framereader.py: fix got float for P210"` commit, and rebased code for commit logs, on my local the commits looks good.
> 
> Did you use following command lines and check locally.
> git fetch origin pull/138/head:pr-138
> git checkout pr-138.
> 
> Anyway, I will continue split commits.

Yes I checked out the latest series.  What I'm saying is that when you rebased (squashed) the two commits above, you combined both _commit messages_ together... you need to drop the message from the "fix" commit.

uartie:
```
commit 143b65d680781f15eeca1679941796a4d2f2724a
Author: Zachary Zhou <zachary.zhou@hotmail.com>
Date:   Tue Dec 3 14:42:10 2019 +0800

    lib/framereader.py: replace / with //
    
    lib/framereader.py: fix got float for P210  <--- PLEASE REMOVE THIS FROM COMMIT MESSAGE
```

zhoujd:
> ```
> commit 143b65d680781f15eeca1679941796a4d2f2724a
> Author: Zachary Zhou <zachary.zhou@hotmail.com>
> Date:   Tue Dec 3 14:42:10 2019 +0800
> 
>     lib/framereader.py: replace / with //
>     
>     lib/framereader.py: fix got float for P210  <--- PLEASE REMOVE THIS FROM COMMIT MESSAGE
> ```

Yes, I have done this change

zhoujd:
Thanks for the review, will merge the patch after Linux images.

zhoujd:
Test passed on all linux distro images with smoke and full test. merging it.

PR #139, opened 2019-11-08
WIP: test/gst: do not print position progress with gst-launch

uartie:
All test command output is redirected to log files.
"Progress" indicators from commands can often generate
very large log files if every progress update is written
to a new line.  This can have a negative impact on disk
storage if log files are archived for future reference
(e.g. CI).

Gstreamer recently added a position progress indicator
to the gst-launch command and enables it by default
(see gstreamer commit 0505bd76a109).  If stdout/err is
a TTY, then the progress indicator isn't an issue since
carriage-returns ('\r') are used to print the progress
(i.e. no linefeeds).  However, in the case of non-TTY
(e.g. log files), each progress update is written to a
new line.  We can use --no-position to disable this
progress indicator.

Thus, add --no-position command-line option to all
gst-launch commands to avoid printing the pipeline
position progress indicator.

### DIFF
--- a/test/gst-msdk/decode/decoder.py
+++ b/test/gst-msdk/decode/decoder.py
@@ -18,7 +18,7 @@ class DecoderTest(slash.Test):
   @timefn("gst")
   def call_gst(self):
     call(
-      "gst-launch-1.0 -vf filesrc location={source}"
+      "gst-launch-1.0 --no-position -vf filesrc location={source}"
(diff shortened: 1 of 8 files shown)


uartie:
Closing for now since gstreamer is considering disabling position progress if non-tty... see https://gitlab.freedesktop.org/gstreamer/gstreamer/merge_requests/319

PR #143, opened 2019-11-25
lib/caps: add CML support

zhoujd:
add caps for CML


zhoujd:
will merge it if no more comments and objection

PR #145, opened 2019-12-04
lib/metrics: skimage verion check for 0.16

zhoujd:


lib/metrics.py
```diff
@@ -121,10 +120,15 @@ def calculate(self, file1, file2, nframes, compare):
     return self.__get()
 
 def __compare_ssim(planes):
+  try:
+    # try skimage >= 0.16, first
+    from skimage.metrics import structural_similarity as skimage_ssim
+  except:
+    from skimage.measure import compare_ssim as skimage_ssim
```
uartie: Please move all of the try/except imports to the top of the file (i.e. outside of the functions)

zhoujd: did the change in new version

zhoujd:
Thanks, will merged

zhoujd:
merged, thanks.

PR #146, opened 2019-12-10
gst-msdk/gst-vaapi:add a format VUYA.

zhangyuankun-star:
We need format "VUYA" to encode and decode some 8bit 444 video 
     stream,so I add it.

Signed-off-by: Yuankun Zhang <yuankunx.zhang@intel.com>


zhangyuankun-star:
For 444 video stream,we should decode and output "ayuv". but in intel, we use "VUYA" to output "ayuv" for gstreamer decode. Now the checksumsink we use in CI do not support "VUYA",but haihao's checksumsink support it .Can we use haihao's checksumsink in CI?

test/gst-msdk/util.py
```diff
@@ -23,7 +23,7 @@ def get_supported_format_map():
     "I420"  : ("i420", "I420"),
     "NV12"  : ("nv12", "NV12"),
     "YV12"  : ("yv12", "YV12"),
-    "AYUV"  : ("ayuv", "AYUV"),
+    "AYUV"  : ("vuya", "VUYA"),
     "YUY2"  : ("yuy2", "YUY2"),
```
xhaihao: I think you should add "VUYA" : ("vuya", "VUYA") here, then replace AYUV with VUYA in the caps files.

uartie:
> Can we use haihao's checksumsink in CI?

Currently we use https://github.com/ceyusa/gst-checksumsink and submit patches there.  If we need patches to merge quicker, then I would recommend we fork checksumsink to https://github.com/intel-media-ci account so we can manage our pull-requests there and have CI use that.  @xhaihao @zhoujd what do you think?

test/gst-msdk/util.py
uartie: agree.

uartie: ayuv mapping should not be removed here.  It may still be useful in other cases.

xhaihao:
> Currently we use https://github.com/ceyusa/gst-checksumsink and submit patches there. If we need patches to merge quicker, then I would recommend we fork checksumsink to https://github.com/intel-media-ci account so we can manage our pull-requests there and have CI use that. @xhaihao @zhoujd what do you think?

I think CI shouldn't depend on any personal repo.  I'm fine with your recommendation.

test/gst-msdk/util.py
zhangyuankun-star: Thanks for your advice,I will modify it soon.

uartie: One more thing... you should also add VUYA to the lib/formats.py subsampling map so tests can add software conversion if needed.

uartie: And to be clear, the caps files specify the supported hardware formats whereas the test/gst-msdk/util.py specify the supported middleware formats.

zhangyuankun-star: I want to add "VUYA" to where exist "AYUV".Can I use this means to deal with this question?

uartie: Consider this example:

A user config can define a test case where the source and/or reference is defined in AYUV format.  When executed, the test function will query the hardware/driver caps file and find only VUYA is supported.  Then, the test will query for a compatible subsampling format (in lib/formats.py) and verify both are supported by middleware.  For raw input, the test will insert a software conversion from AYUV -> VUYA for hardware upload.  For raw outputs, it will insert a software conversion from hardware downloaded VUYA -> AYUV so that it can do the comparison.

uartie:
> > Currently we use https://github.com/ceyusa/gst-checksumsink and submit patches there. If we need patches to merge quicker, then I would recommend we fork checksumsink to https://github.com/intel-media-ci account so we can manage our pull-requests there and have CI use that. @xhaihao @zhoujd what do you think?
> 
> I think CI shouldn't depend on any personal repo. I'm fine with your recommendation.

@zhoujd can you implement this please?

uartie:
If you want to be able to use "VUYA" in user config files (i.e. as source input or reference output), you will need to implement a VUYA frame reader (in framereaders.py) and an entry in the metrics.py::get_framesize.

zhoujd:
> > > Currently we use https://github.com/ceyusa/gst-checksumsink and submit patches there. If we need patches to merge quicker, then I would recommend we fork checksumsink to https://github.com/intel-media-ci account so we can manage our pull-requests there and have CI use that. @xhaihao @zhoujd what do you think?
> > 
> > 
> > I think CI shouldn't depend on any personal repo. I'm fine with your recommendation.
> 
> @zhoujd can you implement this please?

I think it is good, and I will do the change.

test/gst-msdk/util.py
zhangyuankun-star: I have modify my patch to add "VUYA".But I find that I test gst-msdk/gst-vaapi decode by vaapi-fits,the function "read_frame_VUYA" which I add in file "lib/framereader.py " will not execute.I wonder if it is neccessary for us to add it?

uartie: the frame readers are only needed for output metrics.

uartie:
What use case(s) are you trying to target for VUYA?  I don't see any indication in media-driver docs that VUYA is a supported format (input or output) for any codec.

xhaihao:
I guess something was mixed here.

AYUV in gstreamer and AYUV in media-driver are two different formats :( ,

AYUV in media-driver follows Microsoft AYUV definition which is in V-U-Y-A order in memory (https://docs.microsoft.com/en-us/windows/win32/medfound/recommended-8-bit-yuv-formats-for-video-rendering#ayuv),  however AYUV in gstreamer was defined in A-Y-U-V order in memory (https://gstreamer.freedesktop.org/documentation/design/mediatype-video-raw.html#formats, but now this link is broken).  So VUYA was added into gstreamer for Microsoft AYUV this year, hence now VUYA in gstreamer is equal to AYUV in media-driver.

zhangyuankun-star:
In GStreamer, ayuv is defined according to the byte order of memory storage. No matter the big endian or little endian, it is a > y > u > v. Microsoft also defines ayuv, but it is defined according to 4 bytes, so it is v > u > y > a on Intel platform. Intel's hardware and drivers understand ayuv according to this definition. The GStreamer community does not agree to use ayuv to represent Microsoft's ayuv on little endian machines, so haihao added vuya in GStreamer, and associated vuya with Intel's ayuv in GStreamer

uartie:
very confusing

uartie:
What order are the bytes stored on disk (in file)?

xhaihao:
For VUYA, it is V-U-Y-A in byte, and AYUV in 4bytes, which is what we want.  You may run the command below to get a VUYA file,  then check the byte order. By default, A is set to 0xFF. 

$> gst-launch-1.0 videotestsrc num-buffers=1 ! video/x-raw,format=VUYA ! filesink location=output.yuv

uartie:
Do any of the gstreamer vaapi/msdk elements currently work in pipelines with this format?

uartie:
> Do any of the gstreamer vaapi/msdk elements currently work in pipelines with this format?

Nevermind, I just tried hevc 444 8bit decode on ICL and appears to work.

uartie:
@zhangyuankun-star the changes look good except you still need to change the _decode_ _hevc_8_ and _vp9_8_ fmts from AYUV to VUYA in the `lib/caps/ICL/iHD` caps file.

I'm not sure about the ICL encode caps for AYUV/VUYA input formats... @xhaihao?  I could not get gstreamer-vaapi/msdk to encode either format to hevc.

xhaihao:
> I'm not sure about the ICL encode caps for AYUV/VUYA input formats... @xhaihao? I could not get gstreamer-vaapi/msdk to encode either format to hevc.

The input format is VUYA too for gst-msdk/gstreamer-vaapi.  You have to use low-power mode because currently HEVC 8bit 444 encoding is only supported by low power encoding, please refer to https://github.com/intel/media-driver#decodingencoding-features.

uartie:
I tried low-power encode, but it didn't work for me.  I guess I need to patch my kernel.

xhaihao:
You may check HuC status firstly. 

`$> sudo cat /sys/kernel/debug/dri/0/i915_huc_load_status
`

The kernel should be OK if the status is 'RUNNING'.  The command below works for me on ICL.

`gst-launch-1.0 videotestsrc num-buffers=100 ! video/x-raw,format=VUYA ! msdkh265enc low-power=1 ! filesink location=output.hevc`

zhangyuankun-star:
> @zhangyuankun-star the changes look good except you still need to change the _decode_ _hevc_8_ and _vp9_8_ fmts from AYUV to VUYA in the `lib/caps/ICL/iHD` caps file.
> 

Do I need to change from AYUV to VUYA rather than add VUYA to the lib/caps/ICL/iHD caps file?

uartie:
> > @zhangyuankun-star the changes look good except you still need to change the _decode_ _hevc_8_ and _vp9_8_ fmts from AYUV to VUYA in the `lib/caps/ICL/iHD` caps file.
> 
> Do I need to change from AYUV to VUYA rather than add VUYA to the lib/caps/ICL/iHD caps file?

Yes, you need to change it.

zhangyuankun-star:
I have changed it and I think ffmpeg need add VUYA,too.Can I add VUYA for ffmpeg?

uartie:
Can you update the PR title and description.  Also, you can probably split up your 1 commit into multiple smaller commits (e.g. VUYA enabling commit for the framework, a commit for each middleware utils enabling and a caps fix commit).  And based on @xhaihao's response, the encode caps also need changed.

uartie:
sorry it was not clear... also vdenc caps needs changed.

zhoujd:
Forked repo: https://github.com/intel-media-ci/gst-checksumsink
Changed CI to use: https://github.com/intel-media-ci/gst-checksumsink

zhoujd:
The patch looks good to me, will merge it.

zhoujd:
Thanks for the patch, merged

PR #147, opened 2019-12-26
Support smoke/full run under python3

zhoujd:


lib/common.py
```diff
@@ -100,9 +100,15 @@ def call(command, withSlashLogger = True):
     logger = lambda x: None
 
   def readproc(proc):
-    for line in iter(proc.stdout.readline, ''):
-      readproc.output += line
-      logger(line.rstrip('\n'))
+    import sys
+    for line in iter(proc.stdout.readline, b''):
+      # check python version
```
uartie: Why?
What behavior are you seeing when sharing the same code with both versions?

If we create Popen object with universal_newlines=True (i.e. it opens stdout in text mode), will shared code work for both versions?

lib/common.py
```diff
@@ -100,9 +100,15 @@ def call(command, withSlashLogger = True):
...
+      if sys.version_info.major == 2:
+        readproc.output += line
+        logger(line.rstrip('\n'))
+      else:
+        readproc.output += line.decode("utf-8")
+        logger(line.decode("utf-8").rstrip().rstrip('\n'))
```
uartie: This block of code should work for both python 2 and 3, right?

uartie: Hmmm... I guess this block does not always work on both versions.

Can you try to create Popen object with universal_newlines=True parameter (i.e. it will force Popen to use stdout/err in text mode on both versions of python)... and without the readproc changes?

lib/metrics.py
```diff
@@ -8,8 +8,8 @@
 import itertools
 import os
 
-from common import get_media, memoize, timefn
-from framereader import FrameReaders
+from .common import get_media, memoize, timefn
+from .framereader import FrameReaders
```
uartie: It's unfortunate that this (5dd1d67bd81066ba0b7cca92afeba83c03522246) did not get included with #138 PR.  Please separate this commit into separate PR and push immediately since ./vaapi-fits is now broken without it.

uartie:
superseded by #149

PR #150, opened 2019-12-30
ffmpeg-vaapi and ffmpeg-qsv need to change ayuv format to 0yuv format

dryjq:
At present, the decoding format of ffmpeg corresponding to ayuv is 0yuv, which has been verified locally that 0yuv can pass.


uartie:
superseded by #156

PR #153, opened 2020-01-07
Platform ehl

FocusLuo:
To add the vaapi-fits support for platform EHL
Compared to ICL, there are no any VME encode support (AVC/HEVC VME and VP8/MPEG2/MJPEG are not supported)


zhoujd:
Please clean you commit ids, it looks like only one commit id: 
platform: added EHL platform support

is needed.

zhoujd:
Thanks for the patch, merged

PR #154, opened 2020-01-08
use mapRangeInt for int ranges

uartie:
In python3, the '/' operator no longer takes the
range types into account.  Thus, introduce and use
a mapRangeInt method for tests that expect integer
range mapping.


zhoujd:
Thanks for the patch, I will try it locally first

PR #155, opened 2020-01-08
lib/metrics: remove nrmse norm_type parameter

uartie:
The different versions of the skimage nrmse functions
also have different norm_type parameter name.  Both
versions still default to "euclidean" normalization.
Therefore, there is no need to explicitly specify
this parameter.


zhoujd:
Thanks for the patches, I will try it locally, So this patch both works for python2 and python3, right ?

zhoujd:
Thanks for the patch, merged.

PR #156, opened 2020-01-14
To fix/unify 8bit 444 decode support

FocusLuo:
For 8bit 444 format, the iHD named AYUV, the
gstreamer named vuya, and the ffmpeg is naming 0yuv
This patch will follow the AYUV name in the iHD
for test cases, and do mapping for ffmpeg-vaapi/
ffmpeg-qsv and gst-vaapi/msdk to vuya/0yuv on test
 cases implement


uartie:
Hmm...  this is backtracking/reverting from the discussion we had in https://github.com/intel/vaapi-fits/pull/146?  The driver outputs VUYA, so driver caps file should specify VUYA.  The test should insert proper conversion when needed.  That is, if test config file specifies AYUV or 444P format, then the test case will automatically insert a conversion from [driver] VUYA output to AYUV/444P (that is how gstreamer tests work, AFAICT).

@xhaihao to help clarify.

FocusLuo:
Understood, the #146 patch only considered the gstreamer side. This #156 patch is to support both ffmpeg and gstreamer. 
(I had planned to use the vuya name in test cases define in vaapi-fits-full and to use vuya to map to 0yuv for ffmpeg and vuya->vuya for gstreamer, but Zachary suggest to follow the name of ayuv as the iHD side in the Caps/xx/iHD, I though Zachary's suggestion is right, so for to use ffmpeg ayuv->0yuv, and for gstreamer ayuv->vuya )

uartie:
In gstreamer ayuv != vuya and in ffmpeg vuya != 0yuv, so it is strange we would want to map it that way.  In the test case config file, I think it would be best to use AYUV.  Then the tests can do the correct command-line mapping for CSC output.

FocusLuo:
Thanks, we aligned. This patch (with the 148 patch in vaapi-fits-full) are tried to fix the format gap among iHD/FFmpeg/Gstreamer
 to use the AYUV on the test define (vaapi-fits-full, following iHD define), to use 0yuv and vuya for format mapping for correct command-line in ffmpeg and gstreamer

xhaihao:
LGTM.

test/ffmpeg-qsv/util.py
```diff
@@ -46,7 +46,7 @@ def get_supported_format_map():
     "BGRA"  : "bgra",
     "P210"  : "yuv422p10le",
     "P410"  : "yuv444p10le",
-    "AYUV"  : "ayuv",
+    "AYUV"  : "0yuv",
```
uartie: does ffmpeg have a VUYA code?

FocusLuo: In ffmpeg, the 0yuv is the same output data as the gstreamer VUYA output data;
There is no CSC work when to use the 0yuv in ffmpeg and vuya in gstreamer

FocusLuo: There should be a understand issue among the iHD/Gstreamer/FFmpeg
The output are same among the iHD(AYUV), Gstreamer(VUYA) and FFmpeg(0YUV)

uartie: Can you point me to the ffmpeg patches that add 0yuv format?

uartie: From the user test config perspective... user may want to define tests for both AYUV and VUYA byte-order output references.  So, in gstreamer tests, if we are mapping AYUV->VUYA it will make that impossible.

In the iHD driver documentation, they loosely state they support AYUV.  But that is ambiguous because they are actually using the microsoft definition of AYUV, which is in VUYA byte order.  In gstreamer AYUV is not using the microsoft definition... it is actually in AYUV byte order.

Thus, from the test framework perspective we should use "byte-order" based formats and the driver caps file for iHD should be VUYA (not AYUV).

Also, you just indicated that in ffmpeg 0yuv means VUYA...  so we should map VUYA to 0yuv (not AYUV to 0yuv).

uartie: And AFAICT, ffmpeg does not support AYUV (byte-order) format.  So for maximum compatibility, the user config can use VUYA in the test definitions.

Another argument for not mapping AYUV->VUYA in gstreamer tests is that i965 AYUV is not microsoft version, AFAIK.

fulinjie: FYI, some discussion in the community to suggest 0YUV instead of AYUV(since there is no alpha channel support from hardware):

>>RGBA:  R G B A R G B A ...
>>RGB0:  R G B 0 R G B 0 ...
>>RGB24: R G B R G B R G ...
>>
>>Following the same pattern we would have:

>>AYUV:  A Y U V A Y U V ...
>>0YUV:  0 Y U V 0 Y U V ...

>>To me it looks like 0YUV is exactly what you want here.

https://patchwork.ffmpeg.org/project/ffmpeg/patch/1566973334-19107-1-git-send-email-linjie.fu@intel.com/

FocusLuo: > From the user test config perspective... user may want to define tests for both AYUV and VUYA byte-order output references. So, in gstreamer tests, if we are mapping AYUV->VUYA it will make that impossible.
> 
> In the iHD driver documentation, they loosely state they support AYUV. But that is ambiguous because they are actually using the microsoft definition of AYUV, which is in V-U-Y-A byte order. In gstreamer AYUV is not using the microsoft definition... it is actually in A-Y-U-V byte order.
> 
> Thus, from the test framework perspective we should use "byte-order" based format names and the driver caps file for iHD should be VUYA (not AYUV).
> 
> Also, you just indicated that in ffmpeg 0yuv means VUYA... so we should map VUYA to 0yuv (not AYUV to 0yuv).

On the 8bit 444 define among:
iHD: AYUV (following microsoft definition of AYUV, which is in V-U-Y-A byte order on Intel little endian order. It exits many years and maybe it also need to support Microsoft windows OS)
Gstreamer: VUYA (it is byte order define, on Intel little endian platform which is actually in V-U-Y-A byte order)
FFmpeg: 0YUV (it is a double word define, which is actually in V-U-Y-A byte order on Intel little endian platform order)

So on the vaapi-fits-full/vaapi-fits test case name define, is it fine to continue to use the AYUV name to follow Intel iHD announcement? which we are aligned yesterday.
This will keep the same announcement with Intel iHD told to outside (It is also can help to save the explanation to other BU or on JIRA when some others user said iHD is AYUV, but there is no AYUV in vaapi-fits-full/vaapi-fits ),
And more ever, there will have HEVC/VP9 444 8bit encode features on both VME and VDenc test cases to be added soon, it is also used the name of AYUV on iHD side

uartie: > Is it fine to continue to use the AYUV name to follow Intel iHD announcement on the vaapi-fits-full/vaapi-fits test case name define? which we are aligned yesterday.

I think the alignment was on my assumption that 0yuv == ayuv (in byte order) in ffmpeg (but apparently it is not based on what you said) and that we would not map AYUV to VUYA in the tests.

We need to view it from the middleware perspective, since that is what we're ultimately testing.  The middleware's job is to encapsulate/hide the driver details away from the user.

**And the problem is that AYUV has an actual definition in gstreamer as A-Y-U-V byte order.**  So using AYUV at the test config level is ambiguous since you want it to mean something different for ffmpeg (which actually does not have A-Y-U-V byte order output support, from what I gather).  Thus, to eliminate ambiguity at the test config level, we need to normalize for maximum compatibility across middleware.

Consider this... say user wants to use SSIM metric in test case with format=AYUV.  If user has their gstreamer hat on, then they will provide a reference file in A-Y-U-V byte order so it can be compared against the test output.  If we map AYUV to VUYA in the test, then the comparison will be wrong.  Let alone, the test framework's AYUV framereader reads in A-Y-U-V byte order, too.  

On the flip-side, say we decided that AYUV at test framework level actually means V-U-Y-A byte order (and we changed the framereader, too).  Then, this would make it impossible for the user to actually test gstreamer's definition of "AYUV".

Therefore, since ffmpeg does not support any output in A-Y-U-V byte order, we should use VUYA in our test config if we want to share one case definition between all middleware tests.

uartie:
Do you have an _actual working_ ffmpeg patch for 0yuv support?

test/ffmpeg-qsv/util.py
xhaihao: 0yuv==ayuv in ffmpeg except alpha is ignored in 0yuv.

uartie: > 0yuv==ayuv in ffmpeg except alpha is ignored in 0yuv.

Ugh!! This contradicts what Focus said!

uartie: ...here: https://github.com/intel/vaapi-fits/pull/156#discussion_r366672854

xhaihao: Sorry I confused you. I mean in the scope of FFmpeg, 0YUV is same as microsoft AYUV except the alpha channel. Actually FFmpeg doesn't define AYUV pixel format, but it defined AYUV decoder and encoder which follows microsoft AYUV definition. So I think we have to follow microsoft definition too if we want to add AYUV pixel format.

xhaihao: Please see https://github.com/intel/vaapi-fits/pull/156#discussion_r367214962 for why is 0YUV instead of AYUV added in Linjie's patch.

uartie: @fulinjie the patch you linked defines a "ayuv" pix_fmt... do you have an updated patch that changes it to "0yuv" pix_fmt that I can test?

Also, can you provide a ffmpeg command line that can successfully decode 444 hevc 8bit using the "0yuv" as raw output, please?

If all this works as explained, then I'm not convinced that this patch actually produces the same md5 for ffmpeg as gstreamer.  But I need a working patch to test that theory.

fulinjie: @uartie try this repo:
https://github.com/intel-media-ci/ffmpeg-cartwheel/commits/ffmpeg-icl-ehl
with HEVC 444 8 bit decode support for vaapi and qsv with 0yuv:

ffmpeg -hwaccel vaapi -i TSCTX_8bit_RExt_SHARP_1.bin -y vaapi.yuv

Note:
The definition of 0YUV is "0 Y U V",
however when dumping data from driver, it just dumps the data directly from driver(V U Y A actually).

uartie: Thanks @fulinjie!

The repo you provided works and produces the correct md5.  You are right that it dumps the data in V-U-Y-A order.

So, in spite of all my fuss and confusion, I'm willing to accept (with minor reluctance) that we recognize AYUV in the test framework to refer to _microsoft's definition_.  The only caveat is that gstreamer does not have the same meaning for AYUV.  Thus, the trade-off is that we lose the ability to validate gstreamer's definition since we are going to force AYUV to map to VUYA in the gstreamer tests.

In light of this, there are still a few more changes that need to be made in the test framework that I will comment below.

test/gst-msdk/util.py
```diff
@@ -23,8 +23,7 @@ def get_supported_format_map():
     "I420"  : ("i420", "I420"),
     "NV12"  : ("nv12", "NV12"),
     "YV12"  : ("yv12", "YV12"),
-    "AYUV"  : ("ayuv", "AYUV"),
-    "VUYA"  : ("vuya", "VUYA"),
+    "AYUV"  : ("vuya", "VUYA"),
```
uartie: Please put a comment here that states we use microsoft's definition of AYUV (and a link to that definition), please.

test/gst-vaapi/util.py
```diff
@@ -20,8 +20,7 @@ def get_supported_format_map():
     "I420"  : ("i420", "I420"),
     "NV12"  : ("nv12", "NV12"),
     "YV12"  : ("yv12", "YV12"),
-    "AYUV"  : ("ayuv", "AYUV"),
-    "VUYA"  : ("vuya", "VUYA"),
+    "AYUV"  : ("vuya", "VUYA"),
```
uartie: Please put a comment here that states we use microsoft's definition of AYUV (and a link to that definition), please.

test/ffmpeg-qsv/util.py
FocusLuo: Thanks all on align this.
Let me to follow these changes and to give an update soon

test/gst-vaapi/util.py
FocusLuo: updated

test/gst-msdk/util.py
FocusLuo: Sure, updated

uartie:
Thanks @FocusLuo ... LGTM

PR #158, opened 2020-01-17
test/gst-msdk/encode: explicitly specify qpi and qpb in pipeline

uartie:
If qpi and qpb is unspecified, then gst-msdk will
    send 0 to MSDK.  However, MSDK treats 0 as an
    uninitialized value and does not specify the behavior
    for such value.  Thus, MSDK is free to choose/change
    the behavior as it sees fit (which recently happened
    and changed all the test results).
    
    We already explicitly set the qpp in the test.
    
    This adds an explicit setting for qpi and qpb, too.
    Setting all qp's to the same value essentially makes
    it roughly equivalent to gst-vaapi's init-qp
    functionality and should guarantee no surprises if
    MSDK decides to change the behavior for 0 qp's.
    
    Unfortunately, this will require gst-msdk encode
    test baselines to be rebased.

### DIFF
--- a/test/gst-msdk/encode/encoder.py
+++ b/test/gst-msdk/encode/encoder.py
@@ -36,7 +36,7 @@ class EncoderTest(slash.Test):
       if self.codec in ["mpeg2"]:
         opts += " qpi={mqp} qpp={mqp} qpb={mqp}"
       else:
-        opts += " qpp={qp}"
+        opts += " qpi={qp} qpp={qp} qpb={qp}"
     if vars(self).get("quality", None) is not None:
       if self.codec in ["jpeg",]:
         opts += " quality={quality}"
--- a/baseline/default
+++ b/baseline/default
@@ -1,8971 +1,9115 @@
 {
   "test/ffmpeg-qsv/decode/avc.py:default.test(case=1080p)": {
     "md5": "2873526ec49defd754c7853f0658799d"
-  }, 
+  },
   "test/ffmpeg-qsv/decode/avc.py:default.test(case=720p)": {
     "md5": "e6b24a42be6b698f1bf96b75d1480ade"
-  }, 
+  },
   "test/ffmpeg-qsv/decode/avc.py:default.test(case=QCIF)": {
     "md5": "61aeb97a3ab5da00a28a962ae9d69a28"
-  }, 
+  },
   "test/ffmpeg-qsv/decode/avc.py:default.test(case=QVGA)": {
     "md5": "767f88894d9635691c6e1af82b1a738b"
-  }, 
+  },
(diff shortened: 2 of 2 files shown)


test/gst-msdk/encode/encoder.py
```diff
@@ -36,7 +36,7 @@ def gen_output_opts(self):
       if self.codec in ["mpeg2"]:
         opts += " qpi={mqp} qpp={mqp} qpb={mqp}"
       else:
-        opts += " qpp={qp}"
+        opts += " qpi={qp} qpp={qp} qpb={qp}"
```
zhoujd: It looks like the same {qb} for qpi/qpp/qpb.  But from sample_encode output
 [-qpi]                   - constant quantizer for I frames (if bitrace control method is CQP). In range [1,51]. 0 by default, i.e.no limitations on QP.
   [-qpp]                   - constant quantizer for P frames (if bitrace control method is CQP). In range [1,51]. 0 by default, i.e.no limitations on QP.
   [-qpb]                   - constant quantizer for B frames (if bitrace control method is CQP). In range [1,51]. 0 by default, i.e.no limitations on QP.

It looks like qpi/qpp/qpb can be different values. if gst-msdk need they be same ?

xhaihao: no, gst-msdk doesn't need the same value for QPx

uartie: The MSDK library recently changed how QPx=0 is handled, so the sample_encode help is misleading (and ambiguous to begin with).  MSDK used to set QPx=1 when user sent 0, now it sets to different defaults and changes our test results if we aren't explicit.  We are working to get MSDK to make the behavior more clear.  Either way, let's be explicit in the test so we're not at their mercy.  Agree?

It's ok if user explicitly sets QPx to same values, but it is not required.  On ffmpeg side, they use an ipratio/pbratio to set different QPI/QPB relative to QPP.  But in gst-vaapi, they set all QPx the same by default (specified by init-qp prop).  This patch just sets them all explicitly to the same value so gst-msdk can closer match gst-vaapi behavior and to avoid the unspecified behavior by MSDK library.

uartie:
This also fixes another issue in the test.  For test cases with only I-frame encoding (i.e. no P/B-frames), the `qp` parameter from the test config was essentially unused since it was only passed to the qpp property.  And therefore, these aforementioned test cases that only variate the `qp` value always produce the same result and are unnecessarily redundant.  Now, with qpi, qpp and qpb set based on the different specified `qp` from test config, we get some real output variation and meaningful coverage.

uartie:
Merging now... if anyone has strong disagreement we can revert.

PR #160, opened 2020-02-10
test/gst-msdk/encode/vp9.py: Add muxer in command line

xhaihao:
The src caps of msdkvp9enc is video/x-vp9 which means raw VP9 stream, so
a muxer is required if we want to get a stream file.

### DIFF
--- a/test/gst-msdk/encode/vp9.py
+++ b/test/gst-msdk/encode/vp9.py
@@ -15,12 +15,14 @@ class VP9EncoderTest(EncoderTest):
     vars(self).update(
       codec         = "vp9",
       gstencoder    = "msdkvp9enc",
-      gstdecoder    = "ivfparse ! msdkvp9dec hardware=true",
+      gstdecoder    = "matroskademux ! msdkvp9dec hardware=true",
+      gstmediatype  = "video/x-vp9",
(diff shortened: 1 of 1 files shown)


xhaihao:
There is a bug (the scr caps requires video/x-vp9, however the output has an IVF header) in msdkvp9enc. I submitted https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/merge_requests/988 to fix this bug, and I will merge https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/merge_requests/988) soon. To keep consistency with gst-msdk, the command line in vaapi-fits should be changed too.

xhaihao:
I tested it on my ICL without regression.

xhaihao:
No, we needn't set profile=0, msdkvp9enc always output ram VP9 stream with profile=0 for 8bit case. On the other hand, matroskamux doesn't have requirement on profile for video/x-vp9 sink pad.
```
$> gst-inspect-1.0 matroskamux
...
      video/x-vp9
                  width: [ 16, 2147483647 ]
                 height: [ 16, 2147483647 ]
...

```

xhaihao:
BTW, I don't have right to merge PR for vaapi-fits. Could you grant me the permission if possible?

PR #171, opened 2020-03-03
test/gst-vaapi/vpp: don't use deprecated properties

uartie:
The width, height and format properties are being
deprecated.  Use a caps filter instead.

See https://gitlab.freedesktop.org/gstreamer/gstreamer-vaapi/-/merge_requests/289

NOTE: this also removes explicit width and height usage
for deinterlace cases.  Removing those parameters exposes
a possible issue with mpeg2 deinterlace that needs further
triage.


zhoujd:
It looks good to me!

PR #174, opened 2020-03-09
Removed the P210 format support

FocusLuo:
P210 in vaapi-fits are implemented incorrectly as
3 planes (Y plane + U plane + V plane). And the
ffmpeg and gstreamer do not have appropriate formats
to handle unofficial P210 formats

Y210 have better support across the media SW stacks
(drivers & middleware), which had been merged


zhoujd:
It looks good to me.

PR #176, opened 2020-03-11
platform EHL: fixing name of VUYA to AYUV format

FocusLuo:
Just following the other platforms, to rename VUYA
to AYUV for 8bit 444 format


FocusLuo:
When enabled/unified the 8bit 444 format into AYUV, the EHL platform cap enabling is on reviewing status, so missed the VUYA rename to AYUV in the EHL caps/iHD file
This patch is to use AYUV to replace the UVYA on EHL platform caps query as following on other platforms' implement for 8bit 444

PR #179, opened 2020-04-15
vp9_lp enc:enable vp9_lp 10bit encoding for gst-msdk

zhangyuankun-star:
enable vp9 10bit encoding(vdenc) test for gst-msdk.

Signed-off-by: Yuankun Zhang <yuankunx.zhang@intel.com>

### DIFF
--- a/test/gst-msdk/encode/10bit/vp9.py
+++ b/test/gst-msdk/encode/10bit/vp9.py
@@ -1,0 +1,105 @@
+###
+### Copyright (C) 2019-2020 Intel Corporation
+###
+### SPDX-License-Identifier: BSD-3-Clause
+###
+
+from .....lib import *
+from ...util import *
+from ..encoder import EncoderTest
+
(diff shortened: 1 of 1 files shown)


test/gst-msdk/encode/10bit/vp9.py
```diff
@@ -0,0 +1,105 @@
...
+### SPDX-License-Identifier: BSD-3-Clause
+###
+
+from .....lib import *
+from ...util import *
+from ..encoder import EncoderTest
+
+spec = load_test_spec("vp9", "encode", "10bit")
+
+class VP9_10EncoderTest(EncoderTest):
```
zhoujd, line 13: It looks like we already have class VP9_10EncoderTest  in gst-vaapi folder vp9.py ?

uartie: @zhoujd this is not a problem.  It is in a different namespace (module).

zhoujd:
merging it, Thanks for the patch

PR #182, opened 2020-04-21
Added VME "low delay b" support and ffmpeg-vaapi HEVC 8bit 

FocusLuo:
Added need VME "low delay b" feature and ffmpeg-vaapi HEVC 8bit  vme low delay b encode support

### DIFF
--- a/test/ffmpeg-vaapi/encode/encoder.py
+++ b/test/ffmpeg-vaapi/encode/encoder.py
@@ -58,6 +58,10 @@ class EncoderTest(slash.Test):
       opts += " -refs {refs}"
     if vars(self).get("lowpower", None) is not None:
       opts += " -low_power {lowpower}"
+    #"b_strategy", "Strategy to choose between I/P/B-frames" 0->normal, 1->low_delay_b, 2->ref_b
+    #https://patchwork.ffmpeg.org/project/ffmpeg/patch/1586853861-21221-1-git-send-email-linjie.fu@intel.com/
+    if vars(self).get("lowdelayb", None) is not None:
+      opts += " -b_strategy {lowdelayb}"
     if vars(self).get("loopshp", None) is not None:
       opts += " -loop_filter_sharpness {loopshp}"
     if vars(self).get("looplvl", None) is not None:
@@ -154,6 +158,8 @@ class EncoderTest(slash.Test):
 
   def encode(self):
     self.validate_caps()
+    if vars(self).get("lowdelayb", None) is not None:
+      assert type(self.lowdelayb) is int and self.lowdelayb >= 0 and self.lowdelayb <= 1, "invalid lowdelayb value"
 
     iopts = self.gen_input_opts()
     oopts = self.gen_output_opts()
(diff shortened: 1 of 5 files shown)


test/ffmpeg-vaapi/encode/encoder.py
```diff
@@ -212,10 +216,15 @@ def check_output(self):
...
+    if vars(self).get("lowdelayb", None) is not None:
+      assert type(self.lowdelayb) is int and self.lowdelayb >= 0 and self.lowdelayb <= 2, "invalid lowdelayb value"
+      if 1 == self.lowdelayb :
+        ipbmode = 3
     ipbmsgs = [
       "Using intra frames only",
       "Using intra and P-frames",
-      "Using intra, P- and B-frames"
+      "Using intra, P- and B-frames",
+      "delay" #cqp output is "Low delay", cbr/vbr output is "low delay"
```
zhoujd, line 227: a little confuse about the comment, are you saying "cqp" and "cbr/vbr" have the same output ?

FocusLuo: pay attention to that there is caps for "Low" and "low" for ffmpeg output log, so this patch only search  the second word mach

uartie: Since these are passed as regular expressions to `re.match`, then maybe you could use `"[L|l]ow delay"` instead (you may need to check this syntax, though)?

test/ffmpeg-vaapi/encode/encoder.py
```diff
@@ -212,10 +216,15 @@ def check_output(self):
 
     # ipb mode
     ipbmode = 0 if vars(self).get("gop", 0) <= 1 else 1 if vars(self).get("bframes", 0) < 1 else 2
+    if vars(self).get("lowdelayb", None) is not None:
+      assert type(self.lowdelayb) is int and self.lowdelayb >= 0 and self.lowdelayb <= 2, "invalid lowdelayb value"
```
uartie, line 220: Shouldn't this check be done way before attempting to run the pipeline command?

uartie: e.g. right after the call to validate_caps() in encode()

test/ffmpeg-vaapi/encode/hevc.py
```diff
@@ -88,6 +89,30 @@ def test_r2r(self, case, gop, slices, qp, quality, profile):
...
+class cqp_ldb(HEVC8EncoderTest):
+  def init(self, tspec, case, gop, slices, bframes, qp, quality, profile):
+    self.caps = platform.get_caps("vme_lowdelayb", "hevc_8")
+    vars(self).update(tspec[case].copy())
+    vars(self).update(
+      bframes      = bframes,
+      case         = case,
+      gop          = gop,
+      qp           = qp,
+      lowdelayb    = 1,
```
uartie, line 102: If lowdelayb can be 1 or 2, then why not make it a test parameter?

FocusLuo: Artie, the option b_strategy=1 is lowdelayb, but 2 is not a low delay b (2 is another b strategy, Linjie just to follow use the b_strategy in ffmpeg using), the iHD driver only support vme p frame  and vme low delay b frame. 
but no support for the 2.

test/ffmpeg-vaapi/encode/encoder.py
FocusLuo: Ok, I am trying

test/ffmpeg-vaapi/encode/hevc.py
uartie: Ok, so lowdelayb implies a true/false (like lowpower).  We should handle it that way.

test/ffmpeg-vaapi/encode/encoder.py
FocusLuo: yes, updated based on your suggestion

test/ffmpeg-vaapi/encode/encoder.py
FocusLuo: yes, updated based on your suggestion

zhoujd:
one question, how to use low delay B in transcode  test cases ?

zhoujd:
It looks good to me. approving it.

FocusLuo:
> one question, how to use low delay B in transcode test cases ?
This patch is to enable the "low delay b" encode. 
Currently I am doing other patches for transcode, and will supply to review soon

PR #183, opened 2020-04-22
Tests for the ITU conformance bitstreams

uartie:
These test cases are based on the ITU-T H.264(AVC) and H.265(HEVC) conformance
bitstreams available at:

  https://www.itu.int/net/ITU-T/sigdb/spevideo/Hseries-s.htm

To run them,

1. Export VAAPI_FITS_CONFIG_FILE=${VAAPI_FITS_PATH}/config/conformance

2. Export the ITU_T_ASSETS environment variable to specify
the base directory location of your local ITU-T conformance
bitstreams (unzipped).


lib/framereader.py
```diff
@@ -90,6 +90,18 @@ def read_frame_P010(fd, width, height):
 
   return y, uv[0::2].reshape((height2, width2)), uv[1::2].reshape((height2, width2))
 
+def read_frame_I010(fd, width, height):
```
zhoujd, line 93: It looks like I010 is the 10bit version of I420 , right ?
Does iHD driver support this format ?

uartie: Yes, I010 is 10bit version of I420.  It has the same subsampling (4:2:0) and bitdepth (10 bit) as P010.  The only difference between P010 and I010 is the memory layout for each Y, U and V channel (i.e. UV interleaved in one plane vs. U and V in separate planes, respectively).

The HW/drivers don't support I010 format.  But they do support P010.  Thus, lossless SW conversion is possible.  We allow these types of lossless conversions in our tests to accommodate user spec metrics requirements.

More specifically, the pre-determined ITU MD5 references for the 10bit conformance bitstreams are based on I010 output.  The driver only outputs P010... so we have to do a lossless SW conversion to I010 to effectively compare it with the ITU MD5.

zhoujd: Thanks for explanation, It sounds good.
 Regarding "lossless SW conversion to I010 " for iHD,  vaapi-fits do the conversion or other tools can help on the conversion ?

zhoujd:
It looks like this patch trying to let "ITU conformance" as a build-in test cases in vaapi-fits, not in vaapi-fits-smoke etc.

lib/framereader.py
uartie: The conversion is added by vaapi-fits into the middleware command-line

uartie:
> It looks like this patch trying to let "ITU conformance" as a build-in test cases in vaapi-fits, not in vaapi-fits-smoke etc.

Yes.  Is there a question?

PR #196, opened 2020-05-26
ffmpeg-vaapi DI Dynamic mode support

FocusLuo:
to let ffmpeg-vaapi support dynamic mode input from
test cases define

### DIFF
--- a/test/ffmpeg-vaapi/vpp/deinterlace.py
+++ b/test/ffmpeg-vaapi/vpp/deinterlace.py
@@ -23,6 +23,7 @@ class DeinterlaceTest(VppTest):
   _default_modes_ = [
     dict(zip(["method", "rate"], m)) for m in itertools.product(
         _default_methods_, ["field", "frame"])]
+  _test_modes_ = _default_modes_
 
   def before(self):
     vars(self).update(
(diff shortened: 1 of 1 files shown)


test/ffmpeg-vaapi/vpp/deinterlace.py
```diff
@@ -69,13 +70,20 @@ def before(self):
     self.ffdecoder = "h264"
     super(avc, self).before()
 
+  for case, params in spec_avc.items():
+    variants = params.get("rate", None)
+    if params.get("rate", None) is not None:
+      DeinterlaceTest._test_modes_ = [
+        dict(zip(["method", "rate"], m)) for m in itertools.product(
+        DeinterlaceTest._default_methods_, variants)]
+
```
uartie, line 79: This is very awkward.

uartie, line 79: This is very awkward to have this code at class level.  How does updating a class variable for every test affect the slash.parametrize below?

uartie: AFAICT, with this code only the last case (that defines `rate` param) in `spec_avc` will determine the final `_test_modes_` value (of which will be used for all cases during `slash.parametrize`).  This seems very wrong.

Instead, I think it would be better to declare the exact params you want for each case in the user config.  The `gen_vpp_deinterlace_variants` may need to be updated to allow this.

uartie: Possible user config params that we should consider/support in `gen_vpp_deinterlace_variants`:

```python
modes = [dict(rate = "frame"),] # all default methods with "frame" rate.
modes = [dict(method = "bob"),] # "bob" method with all default rates.
modes = [dict(method = "bob", rate = "frame"),] # "bob" method with "frame" rate
modes = [dict(method = "bob", rate = "frame"), dict(method = "weave", rate = "field"] # "bob" method with "frame" rate AND "weave" method with "field" rate.
```

uartie: ...and if no modes is specified, then all default combinations are generated.

uartie:
Can you try #197 please?

FocusLuo:
Let us to use the #197, it is better solution

PR #198, opened 2020-06-11
Caps: added P012 codec format for TGL

FocusLuo:
TGL caps added P012 codec format support for hevc
12bit yuv420 decode and encode


FocusLuo:

PR #199, opened 2020-06-11
vaapi-fits core: added P012 format support in libs

FocusLuo:
added P012 format support in vaapi-fits libs


FocusLuo:

PR #200, opened 2020-06-11
P012 self test and asserts support in vaapi fits

FocusLuo:
root@8f8578038cd1:/home/focus/workspace_focus/vaapi-fits# ./vaapi-fits run test/self/formats.py:test_frame_reader --pl TGL -k fmt=P012                                   9 tests collected
test/self/formats.py .........  PASS
======================================== Session ended. 9 successful, 0 skipped, 0 failed, 0 erroneous. Total duration: 00:00:00 ========================================

root@8f8578038cd1:/home/focus/workspace_focus/vaapi-fits# ./vaapi-fits run test/self/formats.py --pl TGL -k fmt=P012
27 tests collected
test/self/formats.py ...........................  PASS
======================================= Session ended. 27 successful, 0 skipped, 0 failed, 0 erroneous. Total duration: 00:00:02 ========================================

PR #201, opened 2020-06-11
P012 and HEVC 12bit decode implementation in gst msdk

FocusLuo:
Added P012 map and HEVC 12bit decode implementation in gst msdk


uartie:
This PR contains patches from #198, #199 and #200.  Please don't submit the same patch across different PR's.  All of the patches are related, so it should be fine to submit them together in one PR.

FocusLuo:
Thanks for finding this, I will remove the contained commits soon
Because there will another peer PR for "P012 and HEVC 12bit decode implementation in ffmpeg-qsv" just as for gst-msdk,  so it is better to let this in a separate PR as suggestion from Zachary before.

FocusLuo:

PR #206, opened 2020-06-18
test/ffmpeg: disable autoscale for decode tests

uartie:
The autoscale option is added in ffmpeg commit 0b511bd9a548.

For multi resolution clips, we don't want to scale the raw
decoded output since scaling algorithms can produce different
results between various drivers/middleware.  For most codecs,
the decoded result should be the same across different
driver/middleware implementations.  This decoded result can
only be effectively validated and compared when scaling is
disabled.  This also allows us to share a single baseline
reference across all tested drivers/middleware.


fulinjie:
lgtm, thanks @uartie for the quick fix in vaapi-fits.

PR #212, opened 2020-07-06
HEVC 8bit vdenc 444 ayuv cqp and gst-msdk cqp encode support

FocusLuo:
Added HEVC 8bit vdenc 444 ayuv cqp and gst-msdk cqp encode support

### DIFF
--- a/test/gst-msdk/encode/encoder.py
+++ b/test/gst-msdk/encode/encoder.py
@@ -121,8 +121,7 @@ class EncoderTest(slash.Test):
     self.refctx = []
 
   def validate_caps(self):
-    # MSDK does not support AYUV input format even though iHD supports it.
-    ifmts = list(set(self.caps["fmts"]) - set(["AYUV"]))
+    ifmts = self.caps["fmts"]
 
     self.hwformat = map_best_hw_format(self.format, ifmts)
     self.mformat  = mapformat(self.format)
--- a/test/gst-msdk/util.py
+++ b/test/gst-msdk/util.py
@@ -95,6 +95,7 @@ def mapprofile(codec,profile):
     },
     "hevc-8"  : {
       "main"                  : "main",
+      "main444"               : "main-444",
     },
     "hevc-10" : {
       "main10"                : "main-10",


test/gst-msdk/encode/hevc.py
```diff
@@ -89,6 +90,37 @@ def test_r2r(self, case, gop, slices, qp, quality, profile):
     vars(self).setdefault("r2r", 5)
     self.encode()
 
+class cqp_lp_444(HEVC8EncoderTest):
```
uartie, line 93: Why do we need a new class for this?  It is the same as the `cqp_lp` class, just different spec and default profiles.  We should use the `cqp_lp` spec and class, instead, for 444 cases.  The profile can be specified in user spec explicitly.

test/gst-msdk/encode/encoder.py
```diff
@@ -121,8 +121,7 @@ def before(self):
     self.refctx = []
 
   def validate_caps(self):
-    # MSDK does not support AYUV input format even though iHD supports it.
-    ifmts = list(set(self.caps["fmts"]) - set(["AYUV"]))
+    ifmts = list(set(self.caps["fmts"]))
```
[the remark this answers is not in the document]
FocusLuo: Sure, let me update here

test/gst-msdk/encode/hevc.py
FocusLuo: I had wanted to separate the hevc 8bit 444 encode from hevc 8bit encode test result in the test report in future. Just as:
Encode:  hevc
               hevc.lp
               hevc.lp.444
What do you have any suggestion on separate test result in test report?

uartie: You can filter by profile parameter for now.

In the future, I'd like to add support for user-defined categories that allow users to assign cases to arbitrary categories that can be used to filter sets of cases (e.g. "itu-t", "4k", "allegro", "multires", or any arbitrary user-defined category, etc...).  But we have not designed that feature yet.

FocusLuo: Ok, I followed your suggestion and updated the patches

test/gst-msdk/encode/encoder.py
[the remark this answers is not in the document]
FocusLuo: update done

test/gst-msdk/encode/encoder.py
```diff
@@ -121,8 +121,7 @@ def before(self):
     self.refctx = []
 
   def validate_caps(self):
-    # MSDK does not support AYUV input format even though iHD supports it.
-    ifmts = list(set(self.caps["fmts"]) - set(["AYUV"]))
+    ifmts = set(self.caps["fmts"])
```
uartie, line 124: I don't think you need to cast this to a `set()`

PR #213, opened 2020-07-08
gst-msdk: fix hevc low-power type changed

uartie:
The low-power property type changed in upstream for
hevc plugin:

https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/-/merge_requests/1405

Thus, adjust accordingly.


xhaihao:
may we use '1' instead of 'true' and '0' instead of 'false' ? If so, we needn't check the plugin.

uartie:
@xhaihao , yes that might be better since 1|0 is supported as alias for true|false in gstreamer.  I will update the patch.

uartie:
@xhaihao updated.

PR #216, opened 2020-07-21
hevc-8 lp 444: added gst-vaapi profile map

FocusLuo:
added gst-vaapi profile map for hevc 8bit vdenc
444 encode


FocusLuo:
Sample change just as the gst-msdk did.

PR #220, opened 2020-07-28
metrics: Use max possible pixel value for PSNR and SSIM

uartie:
https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio#Definition

PSNR and SSIM should be calculated using the correct max possible
pixel value for each bit depth.  For 8 bit, the skimage uses the
correct default of 255.  But for 10 bit and 12 bit, we read the
pixels into 16/32 bit dtype, and skimage will set the default for
those to the max possible value that the "data type" can handle.
However, the max pixel value is less for 10/12 bit formats. It
should be (2^B)-1.  Thus, specify the data_range param to skimage
PSNR and SSIM so that the correct max possible pixel value is used
in the calculation.

Also, the 10/12 bit frame readers must mask or shift away any
insignificant bits (depending on format) so that the pixel value
starts at the lsb and puts it into the correct pixel value range
for proper metrics calculation.

NOTE: I ran the 10bit HEVC encode smoke and sanity tests on KBL and did not see a huge difference in PSNR results (all passed).  However, this will fix a major issue with Y410 used in encode tests (tests haven't been merged, yet).

### DIFF
--- a/lib/formats.py
+++ b/lib/formats.py
@@ -32,3 +32,6 @@ def match_best_format(fmt, choices):
   if len(matches) == 0:
     return None
   return list(matches)[0]
+
+def get_bit_depth(fmt):
+  return subsampling[fmt][1]
--- a/lib/framereader.py
+++ b/lib/framereader.py
@@ -85,8 +85,8 @@ def read_frame_P010(fd, width, height):
   size    = width * height
   size2   = width2 * height2 * 2
 
-  y = numpy.fromfile(fd, dtype=numpy.uint16, count=size).reshape((height, width))
-  uv = numpy.fromfile(fd, dtype=numpy.uint16, count=size2)
+  y = numpy.fromfile(fd, dtype=numpy.uint16, count=size).reshape((height, width)) >> 6
+  uv = numpy.fromfile(fd, dtype=numpy.uint16, count=size2) >> 6
 
   return y, uv[0::2].reshape((height2, width2)), uv[1::2].reshape((height2, width2))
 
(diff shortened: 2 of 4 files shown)


uartie:
@FocusLuo can you do some extra testing?  Especially for existing P010 encode cases.  And Y210 and Y212 encode cases if any.

FocusLuo:
With this patch and run sanity gst-vaapi on TGL for all encode and transcode, no regression found
test/gst-vaapi/encode/avc.py ................  PASS
test/gst-vaapi/encode/hevc.py sss...sssssssssssssssssssss....ssss......  PASS
test/gst-vaapi/encode/jpeg.py ...  PASS
test/gst-vaapi/encode/mpeg2.py ...  PASS
test/gst-vaapi/encode/vp8.py ssssss  PASS
test/gst-vaapi/encode/vp9.py s..s..s..  PASS
test/gst-vaapi/encode/10bit/hevc.py ssss......sss......sss....  PASS
test/gst-vaapi/transcode/avc.py .s...F  FAIL
test/gst-vaapi/transcode/hevc.py s....  PASS
test/gst-vaapi/transcode/mpeg2.py ..s..  PASS
test/gst-vaapi/transcode/vc1.py .s...  PASS
test/gst-vaapi/vpp/brightness.py .....  PASS
test/gst-vaapi/vpp/contrast.py ...  PASS
test/gst-vaapi/vpp/crop.py ....  PASS
test/gst-vaapi/vpp/csc.py ......  PASS
test/gst-vaapi/vpp/deinterlace.py .s.s  PASS
test/gst-vaapi/vpp/denoise.py ....  PASS
test/gst-vaapi/vpp/hue.py ...  PASS
test/gst-vaapi/vpp/mirroring.py ...  PASS
test/gst-vaapi/vpp/rotation.py ....  PASS
test/gst-vaapi/vpp/saturation.py ...  PASS
test/gst-vaapi/vpp/scale.py ....  PASS
test/gst-vaapi/vpp/sharpen.py ....  PASS
test/gst-vaapi/vpp/transpose.py ......  PASS

PR #224, opened 2020-08-27
bdrate tool

uartie:
tools: add bdrate script
    
    This script calculates bdrate of test results between
    two results.xml files (specified as command line options).
    
    For cqp cases, the 'qp' param is the variant.  For other cases
    (e.g. vbr, cbr), the 'bitrate' param is the variant.
    
    The script depends on the bd-metric python module:
    
      sudo pip3 install bd-metric

cc: @zhangyuankun-star @xuguangxin @FocusLuo

### DIFF
--- a/tools/bdrate
+++ b/tools/bdrate
@@ -1,0 +1,130 @@
+#!/usr/bin/env python3
+
+###
+### Copyright (C) 2020 Intel Corporation
+###
+### SPDX-License-Identifier: BSD-3-Clause
+###
+
+import argparse
+import ast
+import sys
+import xml.etree.cElementTree as et
+
+# HINT: sudo pip3 install bd-metric
+from bd_metric.bjontegaard_metric import *
+
+def parseArgs():
+  parser = argparse.ArgumentParser()
+
+  parser.add_argument(
+    'results',
+    metavar = '<xmlfile>',
+    type = str,
+    nargs = 2,
+    help = "An xml test result file.  The first file is the reference.",
+  )
+
+  parser.add_argument(
+    '-a', '--aggregates',
+    dest = "aggvars",
+    type = str,
+    default = "bitrate,qp",
+    metavar = "VAR[,VAR]",
+    help = "The list of test param names that are aggregated together [default: %(default)s]",
+  )
+
+  parser.add_argument(
+    '--show-points',
+    action = "store_true",
+    help = "Show the (bitrate, psnr) data points used to compute the bdrate",
+  )
+
+  return parser.parse_args()
+
+def aggregate(root):
+  global aggvars
+
+  result = dict()
+  for testcase in root:
+    if testcase.get("skipped", 0) == "1":
+      continue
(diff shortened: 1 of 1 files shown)


uartie:
This gives us many different possibilities for checking bdrate:
 * compare current test results with previous test results (e.g. for regression check)
 * compare test results from one platform to another
 * compare test results from one middleware to another
...and so on.

xuguangxin:
@uartie , you are so efficient.
bdrate calculation is slow, we will selection about 5~10 clips. There are indicate to quality.
If we find a minor psnr change for some test files and these 5~10 clips has no bdrate incremental, we can update the reference psnr.

uartie:
> @uartie , you are so efficient.
> bdrate calculation is slow, we will selection about 5 ~ 10 clips. There are indicate to quality.
> If we find a minor psnr change for some test files and these 5~ 10 clips has no bdrate incremental, we can update the reference psnr.

FWIW, I ran this script on two different HEVCe 10bit results.xml files that I downloaded from our CI Full test runs... it took less than 1 second to parse and compute bd-rate for 132 aggregate cases.

uartie:
I'm going to merge the 2b4b0d7 patch so we can start collecting the bitrate_actual on cqp cases in CI now.

uartie:
> @uartie , you are so efficient.
> bdrate calculation is slow, we will selection about 5 ~ 10 clips. There are indicate to quality.
> If we find a minor psnr change for some test files and these 5~ 10 clips has no bdrate incremental, we can update the reference psnr.

@xuguangxin Can you add those 5 ~ 10 clips to our full test suite?  So, if we see minor psnr regression, then all we have to do is download the results.xml from CI previous and regression run and process them through this script to check bdrate difference.  Then, there is no need to re-run the slow encode pipelines locally.

xuguangxin:
@uartie , which resolution you used, encode and calc psnr for 1080P or 4k is not so fast. 
@zhangyuankun-star will check in the clips and make then run on our full round test everyday.

thanks

uartie:
> @uartie , which resolution you used, encode and calc psnr for 1080P or 4k is not so fast.
> @zhangyuankun-star will check in the clips and make then run on our full round test everyday.
> 
> thanks

@xuguangxin if they are added to full round, then there is no need to encode again since the psnr results are logged in the results.xml files.

uartie:
> > @uartie , which resolution you used, encode and calc psnr for 1080P or 4k is not so fast.
> > @zhangyuankun-star will check in the clips and make then run on our full round test everyday.
> > thanks
> 
> @xuguangxin if they are added to full round, then there is no need to encode again since the psnr results are logged in the results.xml files.

To be clear, this script does not execute any encode and psnr calculation.  It just parses the data that is captured in the results.xml file (which can be downloaded from any finished CI full runs).

xuguangxin:
ok, sorry for misunderstand. 
It's clear to me now, thanks

PR #225, opened 2020-09-03
Ffmpeg vaapi av1 dec 8bit support

FocusLuo:
1, TGL has A stepping and B stepping SKUs,  the B stepping supports the AV1 decode (we have half TGL machines are A stepping)
2, ffmpeg-vaapi support av1 decode


uartie:
How do you propose we handle TGLA vs. TGLB in CI?

I think we should replace TGLA with TGLB (or upgrade them to B-step).  Clearly, TGLA won't be the final product, right?

uartie:
I don't like the idea of adding more complexity here by distinguishing between platform "stepping".  It is better to upgrade our HW to be more representative to the final release product.

FocusLuo:
Yes, TGL-A will

> How do you propose we handle TGLA vs. TGLB in CI?
> 
> I think we should replace TGLA with TGLB (or upgrade them to B-step). Clearly, TGLA won't be the final product, right?

Yes, the TGL-A will be the final product.
I had though to maxium use of these TGL machines in our hands.

FocusLuo:
> I don't like the idea of adding more complexity here by distinguishing between platform "stepping". It is better to upgrade our HW to be more representative of the final release product.
I see, the same thing will be on DG1, most of our DG1 hardware are A stepping board. Very few are B stepping

uartie:
> Yes, TGL-A will
> 
> > How do you propose we handle TGLA vs. TGLB in CI?
> > I think we should replace TGLA with TGLB (or upgrade them to B-step). Clearly, TGLA won't be the final product, right?
> 
> Yes, the TGL-A will be the final product.
> I had though to maxium use of these TGL machines in our hands.

This is a first, AFAIK.  I don't recall that we have ever released an A vs. B vs. ... stepping of a single platform into retail.  Please correct me if I'm wrong. 
 This would change everything if we have platforms under the same codename but with different media capabilities.  I mean, we would have to multiply our HW resources in CI to cover the differences if this is true.  I would like to hear from others in our team on this matter.  @xhaihao @feiwan1, others?  Perhaps we should discuss this further in a team forum.

FocusLuo:
> > Yes, TGL-A will
> > > How do you propose we handle TGLA vs. TGLB in CI?
> > > I think we should replace TGLA with TGLB (or upgrade them to B-step). Clearly, TGLA won't be the final product, right?
> > 
> > 
> > Yes, the TGL-A will be the final product.
> > I had though to maxium use of these TGL machines in our hands.
> 
> This is a first, AFAIK. I don't recall that we have ever released an A vs. B vs. ... stepping of a single platform into retail. Please correct me if I'm wrong.
> This would change everything if we have platforms under the same codename but with different media capabilities. I mean, we would have to multiply our HW resources in CI to cover the differences if this is true. I would like to hear from others in our team on this matter. @xhaihao @feiwan1, others? Perhaps we should discuss this further in a team forum.

My typo, the A stepping will not the final production

uartie:
> > > Yes, TGL-A will
> > > > How do you propose we handle TGLA vs. TGLB in CI?
> > > > I think we should replace TGLA with TGLB (or upgrade them to B-step). Clearly, TGLA won't be the final product, right?
> > > 
> > > 
> > > Yes, the TGL-A will be the final product.
> > > I had though to maxium use of these TGL machines in our hands.
> > 
> > 
> > This is a first, AFAIK. I don't recall that we have ever released an A vs. B vs. ... stepping of a single platform into retail. Please correct me if I'm wrong.
> > This would change everything if we have platforms under the same codename but with different media capabilities. I mean, we would have to multiply our HW resources in CI to cover the differences if this is true. I would like to hear from others in our team on this matter. @xhaihao @feiwan1, others? Perhaps we should discuss this further in a team forum.
> 
> My typo, the A stepping will not the final production

Phew!  We should upgrade our A-step HW to B-step instead of creating special "stepping" caps here in the upstream.

FocusLuo:
> > > Yes, TGL-A will
> > > > How do you propose we handle TGLA vs. TGLB in CI?
> > > > I think we should replace TGLA with TGLB (or upgrade them to B-step). Clearly, TGLA won't be the final product, right?
> > > 
> > > 
> > > Yes, the TGL-A will be the final product.
> > > I had though to maxium use of these TGL machines in our hands.
> > 
> > 
> > This is a first, AFAIK. I don't recall that we have ever released an A vs. B vs. ... stepping of a single platform into retail. Please correct me if I'm wrong.
> > This would change everything if we have platforms under the same codename but with different media capabilities. I mean, we would have to multiply our HW resources in CI to cover the differences if this is true. I would like to hear from others in our team on this matter. @xhaihao @feiwan1, others? Perhaps we should discuss this further in a team forum.
> 
> My typo, the A stepping will not the final production

If you don't like the 

> > > > Yes, TGL-A will
> > > > > How do you propose we handle TGLA vs. TGLB in CI?
> > > > > I think we should replace TGLA with TGLB (or upgrade them to B-step). Clearly, TGLA won't be the final product, right?
> > > > 
> > > > 
> > > > Yes, the TGL-A will be the final product.
> > > > I had though to maxium use of these TGL machines in our hands.
> > > 
> > > 
> > > This is a first, AFAIK. I don't recall that we have ever released an A vs. B vs. ... stepping of a single platform into retail. Please correct me if I'm wrong.
> > > This would change everything if we have platforms under the same codename but with different media capabilities. I mean, we would have to multiply our HW resources in CI to cover the differences if this is true. I would like to hear from others in our team on this matter. @xhaihao @feiwan1, others? Perhaps we should discuss this further in a team forum.
> > 
> > 
> > My typo, the A stepping will not the final production
> 
> Phew! We should upgrade our A-step HW to B-step instead of creating special "stepping" caps here in the upstream.

Fine, let me to update the patch

uartie:
This kind of complexity (i.e. diff stepping) is not currently built into our CI infrastructure.  We need to scope those details if we want to support that.

In the meantime, the caps here (in upstream) should always represent the latest HW iteration of capabilities that we can effectively handle in our CI.  If that means upgrading the HW to the latest stepping (i.e. unreleased HW), then so be it.

uartie (changes requested):
av1_8 should be added to caps/TGL, instead.  And upgrade CI HW to latest stepping.

FocusLuo:
updated patch on caps/TGL

test/ffmpeg-vaapi/decode/av1.py
```diff
@@ -0,0 +1,25 @@
+###
+### Copyright (C) 2018-2019 Intel Corporation
```
uartie, line 2: Copyright (C) 2020 ...

uartie: @FocusLuo please fix this.

FocusLuo: updated

uartie (approved):

PR #227, opened 2020-09-18
Gst vaapi supported hevc 12bit 420 decode

FocusLuo:
gst-vaapi 12bit decode implementation 
P012 format map for gst-vaapi
hevc-12bit 420 profile map for gst-vaapi


uartie (commented):
LGTM

uartie (approved):

PR #231, opened 2020-09-25
ffmpeg-qsv added av1 8bit dec support

FocusLuo:


test/ffmpeg-qsv/decode/av1.py
```diff
@@ -0,0 +1,34 @@
...
+  @slash.requires(*platform.have_caps("decode", "av1_8"))
+  @slash.requires(*have_ffmpeg_decoder("av1_qsv"))
+  @slash.parametrize(("case"), sorted(spec.keys()))
+  def test(self, case):
+    vars(self).update(spec[case].copy())
+    vars(self).update(case = case)
+    self.decode()
+
+  def check_output(self):
+    m = re.search("No support for codec av1", self.output, re.MULTILINE)
```
uartie, line 32: Have you verified that this is the actual warning message emitted by ffmpeg?  In which scenarios will we see this message?

uartie (approved):
LGTM

test/ffmpeg-qsv/decode/av1.py
FocusLuo: @uartie actually, I also didn't see this warning message from ffmpeg-qsv.  Do you think whether I need to remove the check_output function in the av1.py and to give an update the patch again?(noticed that ffmpeg-qsv vp9 also has the check_output function)

for ffmpeg-vaapi, there is a waring message as this in libavcodec/vaapi_decode.c:
root@yami-tgl:/home/focus/workspace_focus/build/ffmpeg-cartwheel# grep -r "No support for codec"
Binary file libavcodec/libavcodec.a matches
libavcodec/vaapi_decode.c:        av_log(avctx, AV_LOG_ERROR, "No support for codec %s "

> Have you verified that this is the actual warning message emitted by ffmpeg? In which scenarios will we see this message?

@uartie actually, I also didn't see this warning message from ffmpeg-qsv.  Do you think whether I need to remove the check_output function in the av1.py and to give an update the patch again?(noticed that ffmpeg-qsv vp9 also has the check_output function)

for ffmpeg-vaapi, there is a waring message as this in libavcodec/vaapi_decode.c:
root@yami-tgl:/home/focus/workspace_focus/build/ffmpeg-cartwheel# grep -r "No support for codec"
Binary file libavcodec/libavcodec.a matches
libavcodec/vaapi_decode.c:        av_log(avctx, AV_LOG_ERROR, "No support for codec

uartie: I guess it's ok to leave it for now since you've shown it is a general decode message... but, eventually we should revisit to make sure which messages we really should look for that indicate hwaccel was skipped and sw fallback used.

uartie (approved):
LGTM

PR #237, opened 2020-10-21
update .slashrc for detect the GPU hang

Bin-QA:
older detect is catch for the 'Resetting' + 'hang' keyword in dmesg,
with current kernel behavior, now just catch GPU hang keyword

Signed-off-by: Bin Wu <binx.wu@intel.com>

### DIFF
--- a/.slashrc
+++ b/.slashrc
@@ -240,8 +240,11 @@ class MediaPlugin(slash.plugins.PluginInterface):
     if slash.config.root.parallel.worker_id is None:
       capture = self.syscapture.checkpoint()
       hangmsgs = [
-        r"\[.*\] i915 0000:00:02.0: Resetting .* after gpu hang",
-        r"\[.*\] i915 0000:00:02.0: Resetting .* for hang on .*",
+        r"\[.*\] i915 .*: \[drm\] Resetting .* after gpu hang",
+        r"\[.*\] i915 .*: \[drm\] Resetting .* for hang on .*",
+        r"\[.*\] i915 .*: \[drm\] .*GPU HANG",
+        r"\[.*\] i915 .*: \[drm\] .*GPU hang",
+        r"\[.*\] i915 .*: \[drm\] Resetting .* time out",
       ]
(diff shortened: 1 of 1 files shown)


Bin-QA:
@uartie I'm not sure here dmesg catch from full dmesg or just for current case  
But I find on TGL and DG1 the keyword already been changed when trigger the GPU Hang  
So we still need use 'Reset' as failed keyword?

.slashrc
```diff
@@ -240,8 +240,9 @@ class MediaPlugin(slash.plugins.PluginInterface):
     if slash.config.root.parallel.worker_id is None:
       capture = self.syscapture.checkpoint()
       hangmsgs = [
-        r"\[.*\] i915 0000:00:02.0: Resetting .* after gpu hang",
-        r"\[.*\] i915 0000:00:02.0: Resetting .* for hang on .*",
+        r"\[.*\] i915 0000:00:0[2-5].0: \[drm\] .*GPU HANG",
+        r"\[.*\] i915 0000:00:0[2-5].0: \[drm\] .*GPU hang",
+        r"\[.*\] i915 0000:00:0[2-5].0: \[drm\] Resetting .* time out",
```
uartie, line 245: The `0000:00:0[2-5].0` regular expression (RE) doesn't make sense to me.  Traditionally, the Intel iGPU is always on `0000:00:02.0`.  And AFAICT, the dGPU (e.g. DG1) is on `0000:03:00.0`, which is not captured by this RE (nor the original RE).

Perhaps `0000:00:02.0` in the RE should be replaced with `.*` instead?

uartie: Also, I think we still want the original RE message, right?  I mean, we want to capture all resets, whether they succeed or not.

Bin-QA: 1. yes, you are right, so we just add new PCI id for dGPU?
2. I forget to check older DUT, which without the kernel upgrade, I will update this PR

Bin-QA: recover form original and remove PCI id detect

PR #239, opened 2020-10-21
ffmpeg-qsv: use '-hwaccel_output_format qsv' for decode

uartie:


uartie:
This makes the command-line compatible before and after this patch: https://github.com/intel-media-ci/ffmpeg/pull/302

PR #244, opened 2020-11-02
ffmpeg-qsv decode vp9 12bit 420 support

FocusLuo:


test/ffmpeg-qsv/decode/12bit/vp9.py
```diff
@@ -0,0 +1,34 @@
+###
+### Copyright (C) 2018-2019 Intel Corporation
```
uartie, line 2: copyright year 2020.

uartie (changes requested):

test/ffmpeg-qsv/decode/12bit/vp9.py
FocusLuo: Updated, thanks reviewing

uartie (approved):
LGTM

PR #246, opened 2020-11-09
lib/gstreamer: use chroma-mode=none for videoconvert

uartie:
Since gst-plugins-base@660b5e4a the chroma-site is
no longer assumed by the base decoder.  This changed
the chroma resampling chosen by videoconvert in many
cases because gst-vaapi and gst-msdk decoders no longer
provide chroma-site setting on their src caps.

Thus, disable chroma resampling on videoconvert to
workaround the issues reported here:

https://gitlab.freedesktop.org/gstreamer/gstreamer-vaapi/-/issues/284
https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/-/issues/1456


FocusLuo:
LGTM

FocusLuo (approved):

PR #248, opened 2020-11-19
Remove low delay B specializations

uartie:
HEVCe LDB is/will be the default for all
vaapi/msdk plugins.

NOTE: HEVCe LDB for ffmpeg-vaapi is only in cartwheel
at the moment.  It doesn't select LDB by default
yet, but a patch will be submitted to enable it as
the default soon.

### DIFF
--- a/lib/caps/DG1/iHD
+++ b/lib/caps/DG1/iHD
@@ -24,8 +24,8 @@ caps = dict(
   encode  = dict(
     avc     = dict(maxres = res4k , fmts = ["NV12"]),
     mpeg2   = dict(maxres = res2k , fmts = ["NV12"]),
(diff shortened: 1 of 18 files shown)


FocusLuo:
@uartie For "VME P frame" path, how will the patch support?

uartie:
> @uartie For "VME P frame" path, how will the patch support?

It shouldn't be related to this patchset.  But, I am experimenting with ideas.  Ideally I would like to avoid adding new test functions, so maybe we will amend test parameters (which can be challenging).

FocusLuo:
Ok, LGTM

uartie:
Thanks @FocusLuo ... I will merge this tomorrow.

PR #251, opened 2020-12-07
gst-msdk add av1 10bit decode support

FocusLuo:
1, TGL caps
2, av1 10bit decode
3, profile map

### DIFF
--- a/lib/caps/TGL/iHD
+++ b/lib/caps/TGL/iHD
@@ -26,6 +26,7 @@ caps = dict(
     vp9_10  = dict(maxres = res8k , fmts = ["P010", "Y410"]),
     vp9_12  = dict(maxres = res8k , fmts = ["P012", "Y412"]),
     av1_8   = dict(maxres = res8k , fmts = ["NV12"]),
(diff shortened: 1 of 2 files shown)


lib/gstreamer/msdk/util.py
```diff
@@ -108,6 +108,9 @@ def mapprofile(codec,profile):
     "av1-8"  : {
       "main"                  : "main",
     },
+    "av1-10"  : {
+      "main"                  : "main-10",
```
uartie, line 112: This should be `"main10" : "main-10"`.

Also, fwiw, `mapprofile` is not used by decoder tests.  It's only used in encoder tests.  Thus, this change is unnecessary for adding decode support in this patchset.

uartie (changes requested):

lib/gstreamer/msdk/util.py
FocusLuo: updated by removing it

uartie (approved):
LGTM

PR #252, opened 2020-12-07
Support test skip for missing features in decoder caps (and define SCC feature)

uartie:
lib: support test skip for missing features in decoders
    
    Some test cases require extra codec features based on platform
    capabilities.
    
    This adds support to decoders to allow test config and
    platform caps to match arbitrary features required for the
    test case to run.  If no match, then the test case can be
    skipped.
    
    For example, in the test case specification the user can define
    "features = ['xyz', 'abc']" to enable the case on platforms
    that only have 'xyz' and 'abc' features defined in the
    platform capabilities file (i.e. in lib/caps/..).
    
    To declare a set of supported/unsupported codec features in the
    platform caps, we can define for example:
    
     "features = dict(xyz = True, abc = True, other = False)"
    
    ...where True means the feature is supported and False means
    the feature is unsupported.  A feature is unsupported by default,
    if the feature is not defined in the dict of features.

    Finally, define scc as supported hevc8 feature on gen12

### DIFF
--- a/lib/__init__.py
+++ b/lib/__init__.py
@@ -11,6 +11,7 @@ from .metrics import *
 from .parameters import *
 from .platform import *
 from .system import *
+from .util import *
 
 import itertools
 import os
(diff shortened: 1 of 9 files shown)


FocusLuo (approved):
LGTM

PR #254, opened 2020-12-10
added ffmpeg-qsv av1-10 decode support

FocusLuo:
ffmpeg-qsv add av1-10 decode support


uartie (approved):
LGTM

PR #255, opened 2020-12-16
gstreamer msdk/vaapi add hevc scc encode support

FocusLuo:
1, Caps: TGL/RKL/DG1
2, gst-msdk add hevc scc encode support
3, gst-vaapi add hevc scc encode support


uartie (approved):
LGTM

PR #256, opened 2020-12-17
ffmpeg-vaapi/qsv hevc scc encode enabling

FocusLuo:
1, ffmpeg-vaapi hevc scc encode enabling
2, ffmpeg-qsv hevc scc encode enabling


FocusLuo:
this implementation same gstreamer-vaapi/msdk

uartie (approved):
LGTM

PR #258, opened 2020-12-20
ffmpeg-qsv/vaapi hevc scc enc 444 enabling

FocusLuo:
ffmpeg-qsv/vaapi hevc scc enc 444 enabling


FocusLuo:
almost same implement as ffmpeg-vaapi/hevc 420 and gst-msdk/vaapi 444

PR #259, opened 2020-12-23
Caps: add sg1 support

FocusLuo:
add sg1 support


FocusLuo:
Only add SG1 caps define in vaapi-fits

PR #260, opened 2020-12-28
gst-vaapi added hevc/vp9 12bit 444 support

FocusLuo:
added hevc/vp9 12bit 444 support


FocusLuo:
Only added "Y412"  : ("y412", "Y412_LE") map for gst-vaapi

PR #261, opened 2020-12-28
Caps: added av1-10 and vp9-12bit decode for sg1/dg1/rkl

FocusLuo:
added av1-10 and vp9-12bit decode for sg1/dg1/rkl


FocusLuo:
Only enabled these features on DG1/SG-1/RKL as the TGL did

PR #262, opened 2020-12-29
gst-msdk added vp9 8bit enc ayuv format support

FocusLuo:
added vp9 8bit enc ayuv format support


FocusLuo:
Change only on excluding ayuv from 
if self.codec not in ["hevc-8", "vp9"]:
       ifmts = list(set(ifmts) - set(["AYUV"]))

PR #265, opened 2021-01-14
gst-vaapi added av1 10bit decode support

FocusLuo:
added av1 10bit decode support

### DIFF
--- a/test/gst-vaapi/decode/10bit/av1.py
+++ b/test/gst-vaapi/decode/10bit/av1.py
@@ -1,0 +1,34 @@
+###
+### Copyright (C) 2018-2021 Intel Corporation
+###
+### SPDX-License-Identifier: BSD-3-Clause
+###
+
+from .....lib import *
+from .....lib.gstreamer.vaapi.util import *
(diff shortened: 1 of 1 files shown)


FocusLuo:
With gst-vaapi av1 decode patch merged and vaapi-fits av1-8bit patch merged, now we can add the 10bit

uartie (approved):

test/gst-vaapi/decode/10bit/av1.py
```diff
@@ -0,0 +1,34 @@
+###
+### Copyright (C) 2018-2019 Intel Corporation
```
uartie, line 2: copyright year should be 2021

uartie (changes requested):

test/gst-vaapi/decode/10bit/av1.py
FocusLuo: updated, thanks

uartie (approved):
Thanks. LGTM

FocusLuo:
Thanks for detail review, yes, I also just came back after launch and find this, you are so fast ~~:)

PR #267, opened 2021-01-15
ffmpeg/vaapi: fix "list" command error caused by WA

uartie:
The gpu gen is only known during "run" command.  Thus, move
LDB workaround to test runtime so we don't break other
commands.


FocusLuo:
LGTM

FocusLuo (approved):

PR #269, opened 2021-01-21
ffmpeg-qsv added hevc-10bit 444 enc profile map

FocusLuo:
added hevc-10bit 444 enc profile map


uartie (approved):
LGTM

PR #272, opened 2021-02-24
gst: use eos-after in checksumsink2

uartie:
Depends on https://github.com/intel-media-ci/gst-checksumsink/pull/1

This allows us to force pipelines to stop processing more
frames when the input stream contains more frames than requested
by the test case.


FocusLuo (approved):
LGTM

PR #273, opened 2021-03-02
User defined caps

uartie:
If VAAPI_FITS_CAPS environment variable is defined, then
    try to load caps from this path.  If no caps found, then
    fallback to library caps path.


uartie:
cc: @Bin-QA

PR #274, opened 2021-03-04
Update the command  for av1 decoding.

fanli-hub:
ivfparse is actually a demuxer, and av1parse is required after ivfparse.

### DIFF
--- a/test/gst-msdk/decode/av1.py
+++ b/test/gst-msdk/decode/av1.py
@@ -29,6 +29,6 @@ class default(DecoderTest):
 
     vars(self).update(
       case        = case,
-      gstdecoder  = "{} ! msdkav1dec".format(dxmap[ext]),
+      gstdecoder  = "{} ! av1parse ! msdkav1dec".format(dxmap[ext]),
     )
     self.decode()
--- a/test/gst-msdk/decode/10bit/av1.py
+++ b/test/gst-msdk/decode/10bit/av1.py
@@ -29,6 +29,6 @@ class default(DecoderTest):
 
     vars(self).update(
       case        = case,
-      gstdecoder  = "{} ! msdkav1dec".format(dxmap[ext]),
+      gstdecoder  = "{} ! av1parse ! msdkav1dec".format(dxmap[ext]),
     )
     self.decode()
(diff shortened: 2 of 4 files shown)


test/gst-msdk/decode/av1.py
```diff
@@ -23,7 +23,7 @@ def before(self):
   def test(self, case):
     vars(self).update(spec[case].copy())
 
-    dxmap = {".ivf" : "ivfparse", ".av1" : "ivfparse", ".webm" : "matroskademux"}
+    dxmap = {".ivf" : "ivfparse", ".av1" : "ivfparse ! av1parse ", ".webm" : "matroskademux"}
```
uartie, line 26: `av1parse` should be added after all demuxers.  Thus, it would be better to just prepend before `msdkav1dec` in the gstdecoder setting on line 32.

uartie (changes requested):
We also need to add av1parse to 10bit test and same for gst-vaapi tests.

xhaihao:
Do we have cases transcoding AV1 to other codecs? If yes, `av1parse` should be added for av1 transcoding too.

uartie:
@fanli-hub please respond to comments

fanli-hub:
We don't have transcode av1.

test/gst-msdk/decode/av1.py
```diff
@@ -23,12 +23,12 @@ def before(self):
   def test(self, case):
     vars(self).update(spec[case].copy())
 
-    dxmap = {".ivf" : "ivfparse", ".av1" : "ivfparse", ".webm" : "matroskademux"}
+    dxmap = {".ivf" : "ivfparse", ".av1" : "ivfparse ", ".webm" : "matroskademux"}
```
uartie, line 26: whitespace change is unnecessary

uartie (changes requested):

uartie (changes requested):
Please squash your commits into one commit.

xhaihao:
@fanli-hub your patch looks good to me, btw will you consider to add transcoding cases for AV1?

uartie:
@fanli-hub your first commit makes a change to dxmap, but the second commit reverts that change.  Thus, please squash both commits into one commit.

PR #283, opened 2021-04-28
Caps: added main still feature support on Gen12

FocusLuo:
added main still picture feature support on Gen12

### DIFF
--- a/lib/caps/ADL/iHD
+++ b/lib/caps/ADL/iHD
@@ -18,7 +18,7 @@ caps = dict(
     hevc_8  = dict(
       maxres    = res8k,
       fmts      = ["NV12", "YUY2", "AYUV"],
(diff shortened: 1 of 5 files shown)


lib/caps/ADL/iHD
```diff
@@ -34,7 +34,7 @@ caps = dict(
     hevc_8  = dict(
       maxres    = res8k,
       fmts      = ["NV12", "AYUV"],
-      features  = dict(scc = True),
+      features  = dict(scc = True, mainstill = True),
```
uartie, line 37: Let's abbreviate `mainstill` as `msp` instead.  You will need to update your previous patches.

FocusLuo: Ok, updated

uartie (approved):
LGTM

PR #287, opened 2021-05-07
slashrc: don't assert device if platform is NONE

uartie:
The unit tests (test/self) are platform agnostic and
can be executed in non-gpu platform (e.g. TravisCI) with
special platform option (--platform NONE).

Thus, if platform == NONE we should not assert that the
gpu device exists.


uartie:
cc: @Bin-CI

PR #289, opened 2021-05-12
test/decode: move requires to class level

uartie:
Previously, in slash < 1.7.10 using slash.requires
at the class level for derived classes caused the
requirements to leak to all sibling classes.  Thus,
the only workaround was to use slash.requires at
each class test method.  This issue was reported in
https://github.com/getslash/slash/issues/928
~2 years ago.

The issue has been fixed since slash 1.7.10.  Now
slash 1.12.0 has been verified to work with vaapi-fits
and is the latest required slash version.

This change moves common slash.requires to the class
level on derived test classes to remove the duplication
that previously occurred across each class test
method.

The slash.requires fix will also enable us to reduce
code duplication by allowing more levels of inheritance
to be introduced so that more code can be shared across
different components.

### DIFF
--- a/test/ffmpeg-qsv/decode/10bit/av1.py
+++ b/test/ffmpeg-qsv/decode/10bit/av1.py
@@ -10,6 +10,8 @@ from .....lib.ffmpeg.qsv.decoder import DecoderTest
 
 spec = load_test_spec("av1", "decode", "10bit")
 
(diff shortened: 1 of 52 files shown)


uartie:
...I will be sending additional, similar patches throughout the week to update encode, vpp and transcode.  The benefits will become more obvious in encode.

PR #291, opened 2021-05-13
test/vpp: move requires to class level

uartie:
This change moves common slash.requires to the class
level on derived test classes to avoid duplicating
at class test methods.

The class-level slash.requires on subclasses are supported
since slash 1.7.10.


uartie:
cc: @Bin-CI

PR #292, opened 2021-05-18
gstreamer: common base classes

uartie:
This is a minimal refactor to consolidate common gstreamer decoder, encoder and vpp test code into base classes so they can be shared between vaapi and msdk (and va) plugin tests without slash.requires and other plugin specific settings leaking across subclasses (e.g. see https://github.com/intel/vaapi-fits/pull/284#pullrequestreview-653824960).  NOTE: this does not include transcoder consolidation (will submit in separate PR).

### DIFF
--- a/lib/gstreamer/decoderbase.py
+++ b/lib/gstreamer/decoderbase.py
@@ -1,0 +1,86 @@
+###
+### Copyright (C) 2018-2021 Intel Corporation
+###
+### SPDX-License-Identifier: BSD-3-Clause
+###
+
+import slash
+
+from ...lib.common import timefn, get_media, call
+from ...lib.formats import match_best_format
+from ...lib.gstreamer.util import have_gst, have_gst_element
+from ...lib.parameters import format_value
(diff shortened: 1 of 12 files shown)


uartie:
cc: @Bin-CI this should be enough to integrate gst-va first-pass.

uartie:
> I think those function also can move into gstreamer as base class function
> 
> ```
> unit.py:
>   def get_supported_format_map
>   mapformat(format):
>   mapformatu(format):
>   map_best_hw_format(format, hwformats):
>   map_transpose_direction(degrees, method):
> encodepy:
>   class EncoderTest(BaseEncoderTest):
>     def map_profile(self):
>     def map_format(self):
>     def map_formatu(self):
> ```

These functions are supposed to return specific values supported by the plugin.  Each plugin does not always have the same supported values.  For example, vaapi plugin might support a new format before the other plugins support it.  They might be the same right now, but this will not always be true.

I will consider a design where common ones can be shared in one place and plugin can override as needed.  This can be done in another PR later.  For now, each plugin should implement their own.

PR #294, opened 2021-05-19
Av1-10 decode added file format obu support for gst-vaapi and gst-msdk

FocusLuo:


test/gst-msdk/decode/10bit/av1.py
```diff
@@ -23,7 +23,7 @@ def before(self):
   def test(self, case):
     vars(self).update(spec[case].copy())
 
-    dxmap = {".ivf" : "ivfparse", ".webm" : "matroskademux"}
+    dxmap = {".ivf" : "ivfparse", ".webm" : "matroskademux", ".obu" : "av1parse"}
```
uartie, line 26: Hmm... we already add `av1parse` unconditionally, below, in `gstdecoder`.  This will produce `gstdecoder = "av1parse ! av1parse ! msdkav1dec"`.  We don't want this.

How about something like:
```python
dxmap = {".ivf" : "ivfparse", ".webm" : "matroskademux", ".obu" : "av1parse"}
ext = os.path.splitext(self.source)[1]
dx = dxmap.get(ext, None)
assert dx is not None, "Unrecognized source file extension {}".format(ext)

if "av1parse" not in dx:
  dx += " ! av1parse"

vars(self).update(
  case             = case,
  gstdecoder  = "{} ! msdkav1dec".format(dx),
)
```

...similar should be done for gst-vaapi/decode/10bit/av1.py

uartie (changes requested):

test/gst-msdk/decode/10bit/av1.py
FocusLuo: updated, thanks

test/gst-vaapi/decode/10bit/av1.py
```diff
@@ -23,12 +23,13 @@ def before(self):
   def test(self, case):
     vars(self).update(spec[case].copy())
 
-    dxmap = {".ivf" : "ivfparse", ".webm" : "matroskademux", ".mkv" : "matroskademux"}
+    dxmap = {".ivf" : "ivfparse", ".webm" : "matroskademux", ".mkv" : "matroskademux", ".obu" : "av1parse"}
     ext = os.path.splitext(self.source)[1]
-    assert ext in dxmap.keys(), "Unrecognized source file extension {}".format(ext)
+    dx = dxmap.get(ext, None)
+    assert dx is not None, "Unrecognized source file extension {}".format(ext)
```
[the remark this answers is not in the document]
FocusLuo: thanks for catching this, updated

uartie (approved):
LGTM

PR #295, opened 2021-05-19
Add hevc-8 mainsp profile map for gst-msdk

FocusLuo:
Add hevc-8 mainsp profile map for gst-msdk


uartie (approved):
LGTM

PR #298, opened 2021-06-02
skip redundant cases with unused params

uartie:
Test function parameters need to be consistent across each
    middleware implemented test.  However, some test functions
    don't support some parameters due to middleware limitations.
    Thus, if a test is variated on the unused parameter(s) it
    creates a redundancy of the same test case.
    
    This change adds support to allow test cases to be skipped
    when unused parameter(s) create a redundancy where the used
    parameters have already been exercised by another case.

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -42,6 +42,23 @@ def timefn(label):
 
   return inner
 
+def parametrize_with_unused(names, values, unused):
+  def inner(func):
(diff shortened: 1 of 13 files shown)


uartie:
Thus supersedes #297

cc: @Bin-CI

FocusLuo (approved):
LGTM

PR #303, opened 2021-06-07
vpp/procamp: normalize range with default/midpoint

uartie:
Each middleware may use a different default/midpoint
    for procamp.  The default point is not always at the
    middle of the range for each procamp operator.
    
    In vaapi-fits, we use the normalized range [0,50,100]
    (i.e. midpoint/default 50).
    
    Thus, take into account the default/midpoint for each
    procamp operator when mapping the value from vaapi-fits
    range into the middleware range.  That is, the vaapi-fits
    default value (50) will map to the middleware default
    value.
    
    This allows for all middleware to share the same procamp
    test case references.
    
    NOTE: this will require any existing procamp test refs
    to be updated.


uartie:
cc: @Bin-CI

uartie:
CI checks depend on https://github.com/intel-innersource/applications.media.middleware.validation.vaapi-fits-smoke/pull/20.
Ran tests manually on KBL and TGL and all cases passed for both drivers and all middleware.

PR #305, opened 2021-06-11
load_test_spec: allow cases to explicitly avoid components

uartie:
Currently, user config cases can be added on a component-specific
basis.  Often, this mechanism is used to avoid running cases on a
specific component.  That is, if the user config wants cases to be
run on all components except component X, then the user config has
to know about all of the components.  This poses a problem where
the user config might forget about one component where the intention
is to run it.  Or, if a new component is added to the framework,
then all user configs need to be updated to ensure cases are added
to that new component.

This change adds a test case property, called "not_components",
that allows the user config to explicitly avoid specific
components for given cases.  Thus, the user config does not
need to know about every available/new component in the
framework.

Example:

```python
media._get_test_spec("jpeg", "encode").update({
  "case_a" : dict(
    format = "AYUV", width = ..., ...,
    not_components = set(["ffmpeg-vaapi",]),
  ),
  ...
})
```

### DIFF
--- a/lib/ffmpeg/qsv/util.py
+++ b/lib/ffmpeg/qsv/util.py
@@ -120,13 +120,5 @@ def mapprofile(codec, profile):
   }.get(codec, {}).get(profile, None)
 
 def load_test_spec(*ctx):
-  from ....lib import get_media
-  import copy
-
-  # get copy of general ctx entries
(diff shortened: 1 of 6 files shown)


FocusLuo:
Ok, we can use the key word of "not_components" to explicitly set not running for some components
it is LGTM

PR #316, opened 2021-07-27
ffmpeg-qsv: add avc forced_idr encode test support

yefengxx:
Add avc forced_idr encode test support for ffmpeg-qsv

### DIFF
--- a/lib/ffmpeg/qsv/encoder.py
+++ b/lib/ffmpeg/qsv/encoder.py
@@ -57,6 +57,8 @@ class EncoderTest(slash.Test):
     if vars(self).get("ladepth", None) is not None:
       opts += " -look_ahead 1"
       opts += " -look_ahead_depth {ladepth}"
+    if vars(self).get("forced_idr", None) is not None:
+      opts += " -force_key_frames expr:1 -forced_idr 1"
 
     opts += " -vframes {frames} -y {encoded}"
 
@@ -90,6 +92,8 @@ class EncoderTest(slash.Test):
       name += "-{lowpower}"
     if vars(self).get("ladepth", None) is not None:
       name += "-{ladepth}"
+    if vars(self).get("forced_idr", None) is not None:
+      name += "-{forced_idr}"
     if vars(self).get("r2r", None) is not None:
       name += "-r2r"
 
@@ -170,6 +174,13 @@ class EncoderTest(slash.Test):
     if vars(self).get("ladepth", None) is not None:
       m = re.search(r"Using the VBR with lookahead \(LA\) ratecontrol method", self.output, re.MULTILINE)
       assert m is not None, "It appears that the lookahead did not load"
+
+    if vars(self).get("forced_idr", None) is not None:
+      output = call(
+        "ffmpeg -v verbose -i {encoded} -c:v copy"
+        " -vframes {frames} -bsf:v trace_headers"
+        " -f null - 2>&1 | grep 'nal_unit_type.*5' | wc -l".format(**vars(self)))
+      assert str(self.frames) == output.strip(), "It appears that the forced_idr did not work"
 
   def check_metrics(self):
     iopts = "-c:v {ffdecoder} -i {encoded}"
--- a/lib/parameters.py
+++ b/lib/parameters.py
@@ -181,6 +181,34 @@ def gen_avc_vbr_la_parameters(spec, profiles):
   params = gen_avc_vbr_la_variants(spec, profiles)
   return keys, params
 
+def gen_avc_forced_idr_variants(spec, profiles):
+  for case, params in spec.items():
+    for variant in copy.deepcopy(params.get("forced_idr", [])):
+      uprofile = variant.get("profile", None)
+      cprofiles = [uprofile] if uprofile else profiles
+      rcmode  = variant["rcmode"]
+      bitrate = variant.get("bitrate", None)
+      qp      = variant.get("qp", None)
+      if "cqp" == rcmode:
+        assert bitrate == None, "We shouldn't set a value to bitrate for CQP."
+        variant.update(maxrate = None)
+      else:
+        assert qp == None, "We shouldn't set a value to qp for CBR/VBR."
+        if "cbr" == rcmode:
+          variant.update(maxrate = bitrate)
+        elif "vbr" == rcmode:
+          variant.update(maxrate = bitrate * 2)
+      for profile in cprofiles:
+        yield [
+          case, rcmode, bitrate, variant["maxrate"],
+          qp, variant["quality"], profile
+        ]
+
+def gen_avc_forced_idr_parameters(spec, profiles):
+  keys = ("case", "rcmode", "bitrate", "maxrate", "qp", "quality", "profile")
+  params = gen_avc_forced_idr_variants(spec, profiles)
+  return keys, params
+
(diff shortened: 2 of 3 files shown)


yefengxx:
@uartie @FocusLuo Can you review this for me? This pr is related to https://github.com/intel-innersource/applications.media.middleware.validation.vaapi-fits-full/pull/27

test/ffmpeg-qsv/encode/avc.py
```diff
@@ -220,3 +220,31 @@ def test(self, case, bframes, bitrate, fps, quality, refs, profile, ladepth):
       ladepth   = ladepth,
     )
     self.encode()
+
+class forced_idr(AVCEncoderTest):
+  def init(self, tspec, case, rcmode, bitrate, qp, quality, profile):
+    vars(self).update(tspec[case].copy())
+    if "cqp" == rcmode:
+        bitrate = None
+        maxrate = None
```
uartie, line 229: User config should not set bitrate or maxrate in this rcmode.  The parameter gen function should set the default to None if not defined in user config, otherwise pass the user config setting to the test function.  Then we can just assert here if they are still not None.

Similar should apply to other rcmode where parameter does not make sense.

uartie (changes requested):

lib/ffmpeg/qsv/encoder.py
```diff
@@ -158,6 +162,10 @@ def encode(self):
         assert result == md5ref, "r2r md5 mismatch"
         # delete encoded file after each iteration
         get_media()._purge_test_artifact(self.encoded)
+    elif vars(self).get("forced_idr", None) is not None:
+      self.check_bitrate()
+      self.check_metrics()
+      self.check_output()
```
uartie, line 168: This `elif` is not needed since the `else` does the same thing.

lib/parameters.py
```diff
@@ -181,6 +181,30 @@ def gen_avc_vbr_la_parameters(spec, profiles):
   params = gen_avc_vbr_la_variants(spec, profiles)
   return keys, params
 
+def gen_avc_forced_idr_variants(spec, profiles, tested):
+  for case, params in spec.items():
+    for variant in copy.deepcopy(params.get("forced_idr", [])):
+      uprofile = variant.get("profile", None)
+      cprofiles = [uprofile] if uprofile else profiles
+      rcmode = variant.get("rcmode", None)
```
uartie, line 189: This implies that `rcmode` might not be required, yet it is unconditionally indexed on line 195.  AFAICT, `rcmode` is required, thus we should use `rcmode = variant["rcmode"]` here, too.  If user config forgets to set it, then it is a config/spec violation and will throw exception.  This is desired because it forces the user to specify it in order for it to be a valid config.  Also, this will allow us to use `rcmode` variable directly on line 195 instead of doing another lookup.

lib/parameters.py
```diff
@@ -181,6 +181,30 @@ def gen_avc_vbr_la_parameters(spec, profiles):
...
+      variant.update(
+        bitrate = None if "cqp" == rcmode else variant["bitrate"],
+        maxrate = None if "cqp" == rcmode else (variant["bitrate"] if "cbr" == rcmode else variant["bitrate"] * 2),
+        qp = None if "cqp" != rcmode else variant["qp"])
```
uartie, line 193: I think it is better to assert this logic instead of silently ignoring/overriding the user test config.  For example, if user test config sets `rcmode = "cqp"` with a `bitrate`, then let it throw an exception because it is an invalid test config definition.

lib/parameters.py
```diff
@@ -181,6 +181,30 @@ def gen_avc_vbr_la_parameters(spec, profiles):
...
+      for profile in cprofiles:
+        testcase = [case, variant["rcmode"], variant["bitrate"],
+          variant["maxrate"], variant["qp"],variant["quality"], profile]
+        #Ensure that there are no duplicate testcase
+        if testcase not in tested:
+          tested.append(testcase)
```
uartie, line 199: This will be unnecessary if we use asserts and exception logic as described.

lib/parameters.py
```diff
@@ -181,6 +181,30 @@ def gen_avc_vbr_la_parameters(spec, profiles):
...
+          yield testcase
+
+def gen_avc_forced_idr_parameters(spec, profiles):
+  keys = ("case", "rcmode", "bitrate", "maxrate", "qp", "quality", "profile")
+  tested = []
+  params = gen_avc_forced_idr_variants(spec, profiles, tested)
```
uartie, line 205: There is no need to pass `tested` as a parameter... it only needs to be a local inside `gen_avc_forced_idr_variants`.  Nonetheless, `tested` goes away with assert/exception logic as described.

uartie (changes requested):

lib/ffmpeg/qsv/encoder.py
```diff
@@ -171,6 +179,17 @@ def check_output(self):
...
+    if vars(self).get("forced_idr", None) is not None:
+      iopts = "-i {encoded} -c:v copy".format(**vars(self))
+      oopts = (
+        "-vframes {frames} -bsf:v trace_headers"
+        " -f null - 2>&1 | grep 'nal_unit_type.*5'"
+        " | wc -l".format(**vars(self)))
+      self.copy_output = call(
+        "ffmpeg -v verbose {iopts} {oopts}".format(iopts = iopts, oopts = oopts))
+      m = re.search(str(self.frames), self.copy_output, re.MULTILINE)
+      assert m is not None, "It appears that the forced_idr did not work"
```
uartie, line 191: Interesting approach... avoids excessive logging messages in original encode command 😉

lib/ffmpeg/qsv/encoder.py
```diff
@@ -171,6 +179,17 @@ def check_output(self):
...
       assert m is not None, "It appears that the lookahead did not load"
 
+    if vars(self).get("forced_idr", None) is not None:
+      iopts = "-i {encoded} -c:v copy".format(**vars(self))
+      oopts = (
+        "-vframes {frames} -bsf:v trace_headers"
+        " -f null - 2>&1 | grep 'nal_unit_type.*5'"
+        " | wc -l".format(**vars(self)))
+      self.copy_output = call(
+        "ffmpeg -v verbose {iopts} {oopts}".format(iopts = iopts, oopts = oopts))
```
uartie, line 189: I suppose this command line can be defined into one variable (e.g. `command`) then just `output = call(command)`.  And, there is no need to store the result in `self` since it is only used locally.

lib/ffmpeg/qsv/encoder.py
uartie, line 191: If the output of the command is only expected to be a single one-liner `number`, then `re` is excessive.  Just do 
```python
assert str(self.frames) == output.strip()
```

lib/ffmpeg/qsv/encoder.py
uartie: Another note, the order here is not good since `check_metrics` can store the PSNR result in baseline if user specifies `--rebase`, but if `check_output` fails next, the rebased metric is useless.

uartie: Either way, just remove this `elif` block and let the `else` do the work in the current correct order.

PR #319, opened 2021-07-28
Validate transcoded resolution

### LINKED ISSUE #119: transcode: verify encoded width/height
In transcode tests, we can do various resolution transcodes.  However, when verifying the transcoded files, we unconditionally decode those files back to the source resolution to do the comparison with the source file.  However, this does not mean the transcoded file was actually transcoded to the requested resolution.  Thus, add an extra check in transcode tests to ensure the transcoded file is in the requested resolution.  For gstreamer tests, we can use gst-discoverer-1.0 -v to check the resolution of the transcoded file.  For ffmpeg tests, we can use ffprobe.

uartie:
Closes #119


uartie:
> LGTM, ffmpeg/qsv and ffmpeg/vaapi seems also have many duplicate code, like older gst

Yep.  I have not started consolidating ffmpeg, yet.  But I am planning to do it.

PR #321, opened 2021-07-29
ffmpeg-vaapi: add vpp composite test

uartie:
overlay_vaapi supported on cartwheel.  vaapi-fits example case passed and internal bat/sanity cases passed on KBL.


uartie:
> Same with #320

Same answers

PR #325, opened 2021-08-10
lib: parse encode config params from variants dict

uartie:
Encode test params should be wrapped inside their
own dict to prevent accidental class variable
conflicts as found in #323.


uartie:
pre-check will fail because it requires smoke config updates.

FocusLuo (approved):

PR #326, opened 2021-08-12
Simplify command line for QSV

xhaihao:
The latest FFmpeg supports hw_device_ctx interface for QSV decoder, we
may use the common device stuff to initialize hw device for QSV
decoder. In addtion, -filter_hw_device option is not needed when there
is only one hw device.


uartie (approved):
LGTM.  Thanks for updating!

PR #338, opened 2021-09-14
gstreamer/encode: use tune property and cqp for avc lp transcode

uartie:
1. The gstreamer msdk elements deprecated the `low-power` property in favor of the `tune` property.
1. AVCe LP only supports CQP rate-control on some older platforms.  Thus, use CQP for AVC LP transcode tests.


FocusLuo (approved):

FocusLuo:
LGTM

PR #350, opened 2021-10-30
ffmpeg-qsv: add hevc forced_idr encode test support

yefengxx:
Add hevc forced_idr encode test support for ffmpeg-qsv

### DIFF
--- a/lib/ffmpeg/encoderbase.py
+++ b/lib/ffmpeg/encoderbase.py
@@ -269,8 +269,11 @@ class BaseEncoderTest(slash.Test):
     if vars(self).get("vforced_idr", None) is None:
       return
 
+    judge = {"hevc-8" : 19, "avc" : 5}.get(self.codec, None)
+    assert judge is not None, f"{self.codec} codec not supported for forced_idr"
+
     output = call(
       "ffmpeg -v verbose -i {encoded} -c:v copy"
       " -vframes {frames} -bsf:v trace_headers"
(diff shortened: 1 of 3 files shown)


lib/ffmpeg/encoderbase.py
```diff
@@ -269,8 +269,13 @@ def check_forced_idr(self):
...
+    if vars(self).get("codec", None) == "hevc-8":
+      judge = " grep 'nal_unit_type.*19' | wc -l"
+    elif vars(self).get("codec", None) == "avc":
+      judge = " grep 'nal_unit_type.*5' | wc -l"
+
     output = call(
       "ffmpeg -v verbose -i {encoded} -c:v copy"
       " -vframes {frames} -bsf:v trace_headers"
-      " -f null - 2>&1 | grep 'nal_unit_type.*5' | wc -l".format(**vars(self)))
+      " -f null - 2>&1 |".format(**vars(self)) + judge)
```
uartie, line 280: `self.codec` is required.  Also, need to fail if no codec matches.  Thus,

```python
judge = {"hevc-8" : 19, "avc" : 5}.get(self.codec, None)
assert judge is not None, f"{self.codec} codec not supported for forced_idr"

...
f" -f null - 2>&1 | grep 'nal_unit_type.*{judge}' | wc -l"
...
```

uartie (changes requested):

PR #351, opened 2021-11-02
[AV1] added gst-msdk av1 cbr lp test implement

FocusLuo:
added gst-msdk av1 cbr lp test implement

Signed-off-by: Focus Luo <focus.luo@intel.com>


uartie (approved):
LGTM

PR #352, opened 2021-11-02
ffmpeg-vaapi added av1 8bit low power cqp and cbr encode support

FocusLuo:
1, profile map
2, lp cqp encode
3, lp cbr encode


uartie (approved):
LGTM

PR #358, opened 2021-11-12
vpp transcoder: use exe2os and filepath2os

tong1wu:

### DIFF
--- a/lib/ffmpeg/transcoderbase.py
+++ b/lib/ffmpeg/transcoderbase.py
@@ -7,7 +7,7 @@
 import re
 import slash
 
-from ...lib.common import timefn, get_media, call
+from ...lib.common import timefn, get_media, call, exe2os, filepath2os
 from ...lib.metrics import calculate_psnr
 from ...lib.ffmpeg.util import have_ffmpeg, ffmpeg_probe_resolution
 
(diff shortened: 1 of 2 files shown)


lib/ffmpeg/transcoderbase.py
```diff
@@ -174,7 +176,8 @@ def check_output(self):
 
   @timefn("ffmpeg")
   def call_ffmpeg(self, iopts, oopts):
-    return call("ffmpeg -v verbose {} {}".format(iopts, oopts))
+    return call(f"{exe2os('ffmpeg')}"
+                " -v verbose {} {}".format(iopts, oopts))
 
   def transcode(self):
     self.validate_caps()
```
uartie, line 183: In the "for loop", below, we dump the transcoded/encoded files to "yuv" files via ffmpeg command.  Thus, we need to use a filepath2os on each yuv file.

Also, in `check_resolution(...)`, we call `ffmpeg_probe_resolution(...)` which will need the encoded/transcoded filename via filepath2os.

uartie (changes requested):

lib/ffmpeg/transcoderbase.py
tong1wu: Updated. Pls review it again, thanks!

uartie (approved):
LGTM

PR #360, opened 2021-12-02
[main10sp ]ffmpeg-qsv added hevc main10sp support

FocusLuo:
1, caps define on Gen12
2, ffmpeg-qsv implement
  ffmpeg -h encoder=hevc_qsv
  -main10sp          <boolean>    E..V....... This profile allow to encode 10 bit single still picture (default false)


lib/caps/DG1/iHD
```diff
@@ -36,7 +36,11 @@ caps = dict(
       fmts      = ["NV12", "AYUV"],
       features  = dict(scc = True, msp = True),
     ),
-    hevc_10 = dict(maxres = res8k , fmts = ["P010", "Y410"]),
+    hevc_10 = dict(
+      maxres = res8k ,
+      fmts = ["P010", "Y410"],
+      features  = dict(m10sp = True)
```
uartie, line 42: 10 bit is already implied in `hevc_10`.  Please just use `msp = True`

lib/ffmpeg/encoderbase.py
```diff
@@ -47,6 +47,7 @@ def gen_output_opts(self):
 
     if vars(self).get("profile", None) is not None:
       opts += " -profile:v {mprofile}"
+      opts += self.gen_hevc_main10sp_opts()
```
uartie, line 50: This function is not defined in base class or other subclasses.  Thus, it will break vaapi, dxva2 and d3d11 encode subclasses.  

How about we return `"main10 -main10sp 1"` from the ffmpeg/qsv/util.py::mapprofile(...) function?  Then `self.mprofile` will already include it and we can remove this new function.

lib/ffmpeg/qsv/encoder.py
```diff
@@ -32,6 +32,10 @@ def gen_quality_opts(self):
       return " -global_quality {quality}"
     return " -preset {quality}"
 
+  def gen_hevc_main10sp_opts(self):
+    if self.profile in ["main10sp"]:
+      return " -main10sp 1"
+
```
uartie, line 38: Just add `-main10sp 1` in mapprofile return value.

lib/ffmpeg/qsv/util.py
```diff
@@ -53,6 +53,7 @@ def mapprofile(codec, profile):
     },
     "hevc-10"  : {
       "main10"     : "main10",
+      "main10sp"   : "main10",
```
uartie, line 56: `"main10sp"   : "main10 -main10sp 1",`

uartie (changes requested):

lib/ffmpeg/encoderbase.py
FocusLuo: sure, thanks Artie

lib/caps/DG1/iHD
FocusLuo: ok

lib/ffmpeg/qsv/util.py
FocusLuo: updated, thanks

lib/ffmpeg/qsv/encoder.py
FocusLuo: updated, thanks

lib/ffmpeg/encoderbase.py
FocusLuo: updated

lib/caps/DG1/iHD
FocusLuo: updated, thanks

uartie (approved):
LGTM

PR #361, opened 2021-12-07
lib/caps/TGL: add vdenc encode cases for libvpl and libmfx

tong1wu:
add vdenc encode cases for libvpl and libmfx


uartie (approved):
LGTM

PR #363, opened 2021-12-24
test/ffmpeg-qsv/encode/avc.py: modify the avc/cbr frames

fanli-hub:
Modify the encode/avc/cbr frames to meet the bitrate_gap requirements.

### DIFF
--- a/test/ffmpeg-qsv/encode/avc.py
+++ b/test/ffmpeg-qsv/encode/avc.py
@@ -96,6 +96,7 @@ class cbr(AVCEncoderTest):
   def init(self, tspec, case, gop, slices, bframes, bitrate, fps, profile):
     vars(self).update(tspec[case].copy())
     vars(self).update(
+      frames    = vars(self).get("brframes", self.frames),
       bframes   = bframes,
       bitrate   = bitrate,
       case      = case,


test/ffmpeg-qsv/encode/avc.py
```diff
@@ -96,6 +96,7 @@ class cbr(AVCEncoderTest):
   def init(self, tspec, case, gop, slices, bframes, bitrate, fps, profile):
     vars(self).update(tspec[case].copy())
     vars(self).update(
+      frames    = vars(self).get("brframes", self.frames),
```
uartie, line 99: Why only support for ffmpeg-qsv cbr?  What about vbr?  What about gst-vaapi, gst-msdk, and ffmpeg-vaapi?

uartie: ...also cbr_lp, vbr_lp, vbr_la, etc.

uartie: Maybe better to add this support to `lib/ffmpeg/encoderbase.py` and `lib/gstreamer/encoderbase.py`?

fanli-hub: I noticed bitrate_gap error only occurred in encode/avc/cbr, not in cbr_lp, vbr, etc. Actually I only paid attention to the ffmpeg-qsv failure cases and didn't notice other frameworks. Following your advice, I checked the other framesworks and observed that the failure cases also existed. So, the vframes need to be modified in other framesworks. I am confused about how to add this support to encoderbase.py? Could you give me some suggestions?

uartie: How about at the beginning of `validate_caps()`, we do:
```python
if self.rcmode in ['cbr', 'vbr', 'la_vbr']:
  # "brframes", if specified, overrides "frames" for bitrate control modes
  self.frames = vars(self).get("brframes", self.frames)
```

uartie: @fanli-hub any update?

fanli-hub: > @fanli-hub any update?

Yes, thank you for your advice. I tried to add this part in encoderbase.py and it worked.  I will push the PR soon.

PR #366, opened 2022-01-06
ffmpeg-d3d11: add av1.py for ffmpeg-d3d11

tong1wu:
add av1.py to enable 8bit av1 test cases

### DIFF
--- a/test/ffmpeg-d3d11/decode/av1.py
+++ b/test/ffmpeg-d3d11/decode/av1.py
@@ -1,0 +1,25 @@
+###
+### Copyright (C) 2022 Intel Corporation
+###
+### SPDX-License-Identifier: BSD-3-Clause
+###
+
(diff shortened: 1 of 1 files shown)


test/ffmpeg-d3d11/decode/av1.py
```diff
@@ -0,0 +1,25 @@
+###
+### Copyright (C) 2021 Intel Corporation
```
uartie, line 2: copyright year should be 2022

uartie (changes requested):

test/ffmpeg-d3d11/decode/av1.py
tong1wu: > copyright year should be 2022

Modified, thanks.

uartie (approved):
LGTM thanks

PR #367, opened 2022-01-06
gst/encoder: take md5 on demuxed video for r2r

uartie:
gstreamer muxers write timestamps to container header and
will be different in each r2r iteration.  Thus, extract the
elementary video stream from the container and take md5 on
the elementary video stream.


FocusLuo (approved):
LGTM

PR #369, opened 2022-01-10
lib/gst/vaapi/encoder: handle minrate

uartie:
gst-vaapi encoders can support custom VBR
    minrate with target-percentage property.
    Thus, set target-percentage property based on
    minrate/maxrate to allow future tests to tweek
    the target.
    
    Currently, gst-vaapi tests hardcode minrate/maxrate
    at 70% which is the gst-vaapi encoder default.  Thus,
    this change should not affect current test case
    results.


uartie:
`target-percentage` property was released in version 1.17... commit https://github.com/GStreamer/gstreamer-vaapi/commit/be496a66c5e8f93849c7c206032c179edc88824b

PR #372, opened 2022-02-02
ffmpeg-vaapi encode: use compression_level option for quality (tu)

uartie:
The base vaapi encoder in ffmpeg uses the global
    -compression_level option to set the libva buffer
    quality level parameter.
    
    Only the vaapi h264 encoder in ffmpeg supports
    the global -quality option to set the libva buffer
    quality level parameter.  However, since the h264
    encoder derives from the base vaapi encoder, it also
    supports the -compression_level option.  Thus, -quality
    is redundant in h264 encoder.
    
    Using -compression_level allows us to properly enable
    the "quality" test parameter usage in the other encoder
    tests (e.g. hevc, vp9...).

### DIFF
--- a/lib/ffmpeg/vaapi/encoder.py
+++ b/lib/ffmpeg/vaapi/encoder.py
@@ -33,7 +33,7 @@ class EncoderTest(BaseEncoderTest):
   def gen_quality_opts(self):
     if self.codec in ["jpeg"]:
       return " -global_quality {quality}"
-    return " -quality {quality}"
+    return " -compression_level {quality}"
 
   def validate_caps(self):
     super().validate_caps()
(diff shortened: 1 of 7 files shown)


uartie:
**FYI:** our internal BAT and SANITY test cases only use default quality/tu level (i.e. 4).  Thus, those will not be impacted by this change and don't need any PSNR rebase.  However, for internal FULL tests we cover additional quality/tu levels.  Thus, we will need to add/rebase MPEG2, VP8, VP9, and HEVC cases (only those which have `quality!=4` variant).  I will work on this.

FocusLuo (approved):
LGTM

PR #374, opened 2022-02-10
metrics: change average range for nrmse

tong1wu:

### DIFF
--- a/lib/metrics.py
+++ b/lib/metrics.py
@@ -292,7 +292,7 @@ def check_metric(**params):
       params["width"], params["height"],
       params["frames"], params["format"])
     get_media()._set_test_details(nrmse = nrmse)
-    avg_range = metric.get("avg_range", [(0, 0.07), (0, 0.07), (0, 0.07)])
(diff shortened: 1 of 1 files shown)


tong1wu:
The nrmse is mainly used for deinterlace. According to @FocusLuo , for deinterlace, what we have to observe is just to see whether there's a regression in the future. Currently, 2 cases on Windows failed because of the 0.07 average range, result in 0.14. So can we just change the number to 0.15 to make the cases pass? Thanks.

uartie:
We should just override the nrmse avg_range in the test config file on a case-by-case basis.  For example:
```python
if media._get_driver_name() in ["d3d11", "dxva2"]:
  v["metric"] = dict(type = "nrmse", avg_range = [(0, 0.15)]*3)
```

tong1wu:
> We should just override the nrmse avg_range in the test config file on a case-by-case basis. For example:
> 
> ```python
> if media._get_driver_name() in ["d3d11", "dxva2"]:
>   v["metric"] = dict(type = "nrmse", avg_range = [(0, 0.15)]*3)
> ```

Got it. I will close this pr and resubmit it in vaapi-fits-smoke

PR #375, opened 2022-02-11
caps: remove av1 test from dxva2 config

tong1wu:
D3D9 does not support av1 decode, so remove them from ci test.


uartie (approved):
LGTM

PR #378, opened 2022-02-18
lib ffmpeg/gst encoderbase:modify the frames

fanli-hub:
Modify the ffmpeg/gst encode frames to meet the bitrate_gap requirements.


fanli-hub:
> the logic abnormal: Father shouldn't know the subclass content and some variable maybe not exist you can check with gen_output_opts style

I have already moved the code from encoderbase.py to ffmpeg/qsv/encoder.py and gst/msdk/encoder.py. Can you help review it?

fanli-hub:
'la_vbr' doesn't have the bitrate_gap issue. So I think we don't need to change the frame in 'la_vbr'.

FocusLuo (approved):
LGTM

PR #379, opened 2022-02-20
vpp/scale: added scale_qsv test for ffmpeg-qsv

yefengxx:
enable scale_qsv test for ffmpeg-qsv, it use

same refs with vpp_qsv test cases.


yefengxx:
@uartie @FocusLuo  can you review this for me ?  And I only modify sanity test(https://github.com/intel-innersource/applications.media.middleware.validation.vaapi-fits-smoke/pull/98) to make sure if you think this is a right way.
Looking forward to your suggestions.

uartie (changes requested):
I think it would be better to create a new test class in a new file `test/ffmpeg-qsv/vpp/scale_qsv.py` and just override gen_output_opts.  The new test class can use same spec as scale.py

yefengxx:
I re-summited a PR in https://github.com/intel/vaapi-fits/pull/391

PR #381, opened 2022-02-22
test/ffmpeg-d3d11/decode: modify ffmpeg-d3d11 test cases

tong1wu:
There are some unsupported formats which are not supported by
ffmpeg-d3d11, now remove them.


tong1wu:
> Can you add some comment for the format description The other guy can quickly understand why need to cleanup the format and add new format

Already add the comments, thx!

test/ffmpeg-d3d11/decode/10bit/hevc.py
```diff
@@ -16,6 +16,9 @@ def before(self):
     # default metric
     self.metric = dict(type = "ssim", miny = 1.0, minu = 1.0, minv = 1.0)
     self.caps   = platform.get_caps("decode", "hevc_10")
+    # ffmpeg-d3d11 only support P010 cases
+    self.caps['fmts'].clear()
+    self.caps['fmts'].append('P010')
```
uartie, line 21: Hmm... I think you may need to do a copy.deepcopy on return value from platform.get_caps to avoid changing other tests that might get the same caps entry.

uartie (changes requested):

test/ffmpeg-d3d11/decode/10bit/hevc.py
uartie: Actuallly, I think it would be better to define a new method in base class that subclass can override to remove unsupported formats.  (see lib/gstreamer/vaapi/vppbase.py for example)

tong1wu: > Actuallly, I think it would be better to define a new method in base class that subclass can override to remove unsupported formats. (see lib/gstreamer/vaapi/vppbase.py for example)

Already updated, pls review the patch thanks!

uartie: To be clear, I mean that the format is not a platform caps limitation, so we should not modify it.  Rather, it is a middleware limitation.  Thus, we should override mapformat and map_best_hw_format which is where middleware specific support formats are selected.

tong1wu: > To be clear, I mean that the format is not a platform caps limitation, so we should not modify the caps. Rather, it is a middleware limitation. Thus, we should override mapformat and map_best_hw_format which is where middleware specific supported formats are selected.

updated, thanks.

test/ffmpeg-d3d11/decode/hevc.py
```diff
@@ -21,5 +21,7 @@ def before(self):
   @slash.parametrize(("case"), sorted(spec.keys()))
   def test(self, case):
     vars(self).update(spec[case].copy())
+    if spec[case].get('features', {}) and spec[case].get('features', {})[0] == 'scc':
```
uartie, line 24: The "features" is a list.  Thus, default return value should be a list `[]`, not a dict `{}`.
How about simplify this to:
```python
if "scc" in vars(self).get("features", list()):
  ...
```

lib/ffmpeg/decoderbase.py
```diff
@@ -43,9 +43,15 @@ def gen_name(self):
       name += "_r2r"
     return name
 
+  def map_best_hw_format(self):
+    return map_best_hw_format(self.format, self.caps["fmts"])
+
+  def mapformat(self):
+    return mapformat(self.format)
+
```
uartie, line 51: I guess this creates the same issue as discussed in #386.  Thus, we better wait for the design fix in #388 to ensure we can override it properly here too.

test/ffmpeg-dxva2/decode/10bit/hevc.py
```diff
@@ -21,5 +21,7 @@ def before(self):
   @slash.parametrize(("case"), sorted(spec.keys()))
   def test(self, case):
     vars(self).update(spec[case].copy())
+    if spec[case].get('features', {}) and spec[case].get('features', {})[0] == 'scc':
```
uartie, line 24: ditto.

uartie (changes requested):

lib/ffmpeg/d3d11/decoder.py
```diff
@@ -15,3 +15,10 @@ class DecoderTest(BaseDecoderTest):
   def before(self):
     super().before()
     self.hwaccel = "d3d11va"
+
+  def mapformat(self):
+    return {
+      "I420"  : "yuv420p",
+      "NV12"  : "nv12",
+      "P010"  : "p010le",
+    }.get(self.format, None)
```
uartie, line 24: Now just override [get_supported_format_map](https://github.com/intel/vaapi-fits/blob/2c4ca12de2f39d3f4faa93dab0048975f398f97f/lib/formats.py#L42) here, instead.

lib/ffmpeg/dxva2/decoder.py
```diff
@@ -15,3 +15,10 @@ class DecoderTest(BaseDecoderTest):
   def before(self):
     super().before()
     self.hwaccel = "dxva2"
+
+  def mapformat(self):
+    return {
+      "I420"  : "yuv420p",
+      "NV12"  : "nv12",
+      "P010"  : "p010le",
+    }.get(self.format, None)
```
uartie, line 24: Now just override [get_supported_format_map](https://github.com/intel/vaapi-fits/blob/2c4ca12de2f39d3f4faa93dab0048975f398f97f/lib/formats.py#L42) here, instead.

uartie:
commit message does not fully represent the actual changes.  please update commit message.

uartie:
Commit message only explains d3d11 format change.  But you've also changed dxva2 formats.  And you've added scc checks.  And you've deleted jpeg decode tests.  Please split changes into logical commits and explain each change.

tong1wu:
> Commit message only explains d3d11 format change. But you've also changed dxva2 formats. And you've added scc checks. And you've deleted jpeg decode tests. Please split changes into logical commits and explain each change.

ok, I will update, thanks.

uartie (approved):

PR #384, opened 2022-03-04
lib/gstreamer/msdk/vpp:remove scaling-mode parameter

fanli-hub:
Align the gst-msdk cmd arguments with ffmpeg-qsv.


mengker33:
The parameter scaling-mode is set as 1 in gst-msdk vpp scaling cmd line, which refers to low-power mode, some cases fail due to ssim value < 0.97 and I suspect setting this param may have an impact on this. Also, FFmpeg-qsv does not set scaling-mode as low-power explicitly,  and gst-msdk failed cases pass in FFmpeg-qsv. So it makes sense to not set scaling-mode in gst-msdk vpp scaling cmd line as well as make cmd align with FFmpeg-qsv.

PR #394, opened 2022-04-22
lib/ffmpeg/encoderbase: add Intra refresh cases

fanli-hub:
Add Intra refresh cases to lib/ffmpeg/encoderbase.

### DIFF
--- a/lib/ffmpeg/encoderbase.py
+++ b/lib/ffmpeg/encoderbase.py
@@ -86,6 +86,10 @@ class BaseEncoderTest(slash.Test, BaseFormatMapper):
       opts += " -look_ahead_depth {ladepth}"
     if vars(self).get("vforced_idr", None) is not None:
       opts += " -force_key_frames expr:1 -forced_idr 1"
+    if vars(self).get("intref", None) is not None:
+      opts += " -int_ref_type {intref[type]}"
+      opts += " -int_ref_cycle_size {intref[size]}"
+      opts += " -int_ref_cycle_dist {intref[dist]}"
 
     # WA: LDB is not enabled by default for HEVCe on gen11+, yet.
     if get_media()._get_gpu_gen() >= 11 and self.codec.startswith("hevc"):
--- a/test/ffmpeg-qsv/encode/avc.py
+++ b/test/ffmpeg-qsv/encode/avc.py
@@ -240,3 +240,45 @@ class forced_idr(AVCEncoderTest):
   def test(self, case, rcmode, bitrate, maxrate, qp, quality, profile):
     self.init(spec, case, rcmode, bitrate, maxrate, qp, quality, profile)
     self.encode()
+
+class intref(AVCEncoderTest):
+  def init(self, tspec, case, gop, bframes, bitrate, qp, maxrate, profile, rcmode, reftype, refsize, refdist):
+    vars(self).update(tspec[case].copy())
+    vars(self).update(
+      rcmode     = rcmode,
+      bframes    = bframes,
+      bitrate    = bitrate,
+      case       = case,
+      maxrate    = maxrate,
+      minrate    = bitrate,
+      profile    = profile,
+      qp         = qp,
+      gop        = gop,
+      intref     = dict(type = reftype, size = refsize, dist = refdist),
+    )
+
+  @slash.parametrize(*gen_avc_intref_parameters(spec, ['high', 'main', 'baseline']))
+  def test(self, case, gop, bframes, bitrate, qp, maxrate, profile, rcmode, reftype, refsize, refdist):
+    self.init(spec, case, gop, bframes, bitrate, qp, maxrate, profile, rcmode, reftype, refsize, refdist)
+    self.encode()
+
+class intref_lp(AVCEncoderLPTest):
+  def init(self, tspec, case, gop, bframes, bitrate, qp, maxrate, profile, rcmode, reftype, refsize, refdist):
+    vars(self).update(tspec[case].copy())
+    vars(self).update(
+      rcmode     = rcmode,
+      bframes    = bframes,
+      bitrate    = bitrate,
+      case       = case,
+      maxrate    = maxrate,
+      minrate    = bitrate,
+      profile    = profile,
+      qp         = qp,
+      gop        = gop,
+      intref     = dict(type = reftype, size = refsize, dist = refdist),
+    )
+
+  @slash.parametrize(*gen_avc_intref_lp_parameters(spec, ['high', 'main', 'baseline']))
+  def test(self, case, gop, bframes, bitrate, qp, maxrate, profile, rcmode, reftype, refsize, refdist):
+    self.init(spec, case, gop, bframes, bitrate, qp, maxrate, profile, rcmode, reftype, refsize, refdist)
+    self.encode()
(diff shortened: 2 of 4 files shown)


fanli-hub:
> CI failed result for GST not this part change cause

Mengke is helping to add this part into gst. And I will add the parameter to gst cmd after Mengke completes it.

uartie:
where are the tests that use this feature?

fanli-hub:
"intra refresh" works when gop > int_ref_cycle_dist > int_ref_cycle_size. The value of cqp doesn't match the rule.

lib/ffmpeg/encoderbase.py
```diff
@@ -87,6 +87,9 @@ def gen_output_opts(self):
     if vars(self).get("vforced_idr", None) is not None:
       opts += " -force_key_frames expr:1 -forced_idr 1"
 
+    if self.codec.startswith("avc") and self.rcmode in ["cbr", "vbr", "cbr_lp", "vbr_lp"]:
+      opts += " -int_ref_type 1 -int_ref_cycle_size 4 -int_ref_cycle_dist 20"
+
```
uartie, line 92: 1. There is no `cbr_lp` or `vbr_lp` for `self.rcmode`.  The lp is only captured in `self.lowpower`.  Thus, only need to check for `cbr` and `vbr` if want to apply to both vme and vdenc.
2. These properties are only valid for ffmpeg-qsv.  This adds them for all ffmpeg plugins, which is not correct.
3. Both AVC and HEVC support these properties in ffmpeg-qsv.  So why only AVC enabled?
4. This is essentially the same hack attempted earlier in the config file patch, just in another disguise.  As explained there, this adds these (hard-coded) parameters for all test cases in these categories.  Thus, this changes pre-existing cases and invalidates everyone's baseline refs for all test suites (quality, sanity, bat, full,...).  This is not acceptable.

Please create a new test for intra refresh and configure separate cases for them in the test configs.  Look at `test/ffmpeg-qsv/encode/avc.py::forced_idr` test for an example.  And the test function should include the intra refresh parameters that will be configured by user test config/spec.

uartie (changes requested):
unacceptable

lib/ffmpeg/encoderbase.py
```diff
@@ -86,6 +86,8 @@ def gen_output_opts(self):
       opts += " -look_ahead_depth {ladepth}"
     if vars(self).get("vforced_idr", None) is not None:
       opts += " -force_key_frames expr:1 -forced_idr 1"
+    if vars(self).get("vintr_ref", None) is not None:
+      opts += " -int_ref_type 1 -int_ref_cycle_size 4 -int_ref_cycle_dist 20"
```
uartie, line 90: Why hardcoded values?  Maybe better to let test config set these parameter values.

uartie, line 90: Why hardcoded values?  Maybe better to let test config set these parameter values.

fanli-hub: > Why hardcoded values? Maybe better to let test config set these parameter values.

How about doing like this time?
Moreover, hevc "-low_power 0" doesn't support Intra refresh, so I only add intra refresh when hevc "-low_power 1".

lib/ffmpeg/encoderbase.py
```diff
@@ -86,6 +86,12 @@ def gen_output_opts(self):
       opts += " -look_ahead_depth {ladepth}"
     if vars(self).get("vforced_idr", None) is not None:
       opts += " -force_key_frames expr:1 -forced_idr 1"
+    if vars(self).get("ref_type", None) is not None:
+      opts += " -int_ref_type {ref_type}"
+    if vars(self).get("ref_type", None) is not None:
+      opts += " -int_ref_cycle_size {ref_size}"
+    if vars(self).get("ref_type", None) is not None:
+      opts += " -int_ref_cycle_dist {ref_dist}"
```
uartie, line 94: copy/paste error... all these conditionals are exactly the same.

The intra refresh type, size, and dist depend on each other.  That is, it does not make sense to have one but not the others.  Therefore, how about a variable like `intref = dict(type = T, size = S, dist = D)`.  Then do:

```python
if vars(self).get("intref", None) is not None:
  opts += " -int_ref_type {intref[type]}"
  opts += " -int_ref_cycle_size {intref[size]}"
  opts += " -int_ref_cycle_dist {intref[dist]}"
```

test/ffmpeg-qsv/encode/avc.py
```diff
@@ -240,3 +240,49 @@ def init(self, tspec, case, rcmode, bitrate, maxrate, qp, quality, profile):
   def test(self, case, rcmode, bitrate, maxrate, qp, quality, profile):
     self.init(spec, case, rcmode, bitrate, maxrate, qp, quality, profile)
     self.encode()
+
+class intr_ref(AVCEncoderTest):
```
uartie, line 244: How about `class intref` instead?

test/ffmpeg-qsv/encode/avc.py
```diff
@@ -240,3 +240,49 @@ def init(self, tspec, case, rcmode, bitrate, maxrate, qp, quality, profile):
...
+  def init(self, tspec, case, gop, bframes, bitrate, qp, maxrate, profile, rcmode, ref_type, ref_size, ref_dist):
```
uartie, line 245: How about `reftype, refsize, refdist` param names instead?

test/ffmpeg-qsv/encode/avc.py
```diff
@@ -240,3 +240,49 @@ def init(self, tspec, case, rcmode, bitrate, maxrate, qp, quality, profile):
...
+      ref_size   = ref_size,
+      ref_dist   = ref_dist,
+    )
+
+  @slash.parametrize(*gen_avc_intr_ref_parameters(spec, ['high', 'main', 'baseline']))
+  def test(self, case, gop, bframes, bitrate, qp, maxrate, profile, rcmode, ref_type, ref_size, ref_dist):
+    self.init(spec, case, gop, bframes, bitrate, qp, maxrate, profile, rcmode, ref_type, ref_size, ref_dist)
+    self.encode()
+
+class intr_ref_lp(AVCEncoderLPTest):
```
uartie, line 267: `class intref_lp`

test/ffmpeg-qsv/encode/avc.py
```diff
@@ -240,3 +240,49 @@ def init(self, tspec, case, rcmode, bitrate, maxrate, qp, quality, profile):
...
+      bitrate    = bitrate,
+      case       = case,
+      maxrate    = maxrate,
+      minrate    = bitrate,
+      profile    = profile,
+      qp         = qp,
+      gop        = gop,
+      ref_type   = ref_type,
+      ref_size   = ref_size,
+      ref_dist   = ref_dist,
```
uartie, line 259: With other suggestions, do:
```python
   intref = dict(type = reftype, size = refsize, dist = refdist)
```

uartie (changes requested):
almost there ;)

lib/ffmpeg/encoderbase.py
uartie: ok, good.

test/ffmpeg-qsv/encode/avc.py
fanli-hub: Thank you for your patience and detailed reply. (●'◡'●)

PR #403, opened 2022-05-23
fmpeg-qsv: remove b_strategy from cmd-line for quality

yefengxx:
remove b_strategy from cmd-line for ffmpeg-qsv vbr_default

### DIFF
--- a/lib/ffmpeg/encoderbase.py
+++ b/lib/ffmpeg/encoderbase.py
@@ -69,7 +69,8 @@ class BaseEncoderTest(slash.Test, BaseFormatMapper):
     if vars(self).get("minrate", None) is not None:
       opts += " -b:v {minrate}k"
     if vars(self).get("maxrate", None) is not None:
-      opts += " -maxrate {maxrate}k"
+      if "vbr_default" != self.rcmode:
+        opts += " -maxrate {maxrate}k"
     if vars(self).get("refs", None) is not None:
       opts += " -refs {refs}"
     if vars(self).get("lowpower", None) is not None:
@@ -89,7 +90,8 @@ class BaseEncoderTest(slash.Test, BaseFormatMapper):
 
     # WA: LDB is not enabled by default for HEVCe on gen11+, yet.
     if get_media()._get_gpu_gen() >= 11 and self.codec.startswith("hevc"):
-      opts += " -b_strategy 1"
+      if vars(self).get("default", None) is None:
+        opts += " -b_strategy 1"
 
     opts += " -vframes {frames} -y {osencoded}"
 


lib/ffmpeg/encoderbase.py
```diff
@@ -69,7 +69,8 @@ def gen_output_opts(self):
     if vars(self).get("minrate", None) is not None:
       opts += " -b:v {minrate}k"
     if vars(self).get("maxrate", None) is not None:
-      opts += " -maxrate {maxrate}k"
+      if "vbr_default" != self.rcmode:
+        opts += " -maxrate {maxrate}k"
```
uartie, line 73: Please don't introduce non-standard or custom rcmode values.

Also, IIUC, if `-maxrate` is not specified then ffmpeg just sets it to target rate (e.g. bitrate)... which is actually CBR, not VBR.

lib/ffmpeg/encoderbase.py
```diff
@@ -89,7 +90,8 @@ def gen_output_opts(self):
 
     # WA: LDB is not enabled by default for HEVCe on gen11+, yet.
     if get_media()._get_gpu_gen() >= 11 and self.codec.startswith("hevc"):
-      opts += " -b_strategy 1"
+      if vars(self).get("default", None) is None:
+        opts += " -b_strategy 1"
```
uartie, line 94: Hmmm... introducing a workaround for another workaround.  Also, a variable called `default` is a bit too broad in meaning.  Clearly we should be able to find a better approach than this.

uartie, line 94: Hmmm... introducing a workaround for another workaround.  Clearly we can take a better approach than this.  Also, a variable called `default` is a bit too broad in meaning.

uartie (changes requested):
Very awkward workarounds...

Please provide us with a high level documentation/test plan so we can better understand what is trying to be achieved here.  What is the end goal and what is missing in the current test framework that is preventing from this goal (or why it seems necessary to add these kind of workarounds)?

@xuguangxin @FocusLuo please help advise.

lib/ffmpeg/encoderbase.py
uartie: > Also, IIUC, if `-maxrate` is not specified then ffmpeg just sets it to target rate (e.g. bitrate)... which is actually CBR, not VBR.

Actually it's different behavior between ffmpeg-qsv and ffmpeg-vaapi.  So it's better to be explicit with both min and max rates.

lib/ffmpeg/encoderbase.py
yefengxx: Can you give me some advice?

uartie: How about:

```
if get_media()._get_gpu_gen() >= 11 and self.codec.startswith("hevc"):
  if vars(self).get("wa_ldb", True):
    opts += " -b_strategy 1"
```

...then in your custom test config file, add `wa_ldb = False` to the test case definition.

lib/ffmpeg/encoderbase.py
uartie: For maxrate, it is hard-coded as a function of bitrate in most/all tests.  You can change the test function in the "quality" suite to use the dict `setdefault` instead of `update`.  This way, if it's specified in user config then it will be used, else the hard-coded default function of bitrate will be used.

```python
vars(self).update(minrate = self.bitrate)
vars(self).setdefault("maxrate", self.bitrate * 2)
```

...then in user config, you can set `maxrate = None` in test case definition.  This is still a little awkward and a bit of a hack, but probably more obvious to anyone that reads it later on.

PR #404, opened 2022-05-24
Some fixups for intra refresh tests

uartie:
1. Fix default profiles in hevce
2. Validate/check for intra refresh command output messages
3. Ensure intref params are reflected in the name to minimize duplicate with other test case artifact names.


uartie:
cc: @fanli-hub

PR #407, opened 2022-05-30
caps: remove ayuv from enc/jpeg

fanli-hub:
Media driver does not support jpeg/ayuv anymore.


FocusLuo (approved):
Based on iHD, there is no AYUV input format supported 
https://github.com/intel/media-driver/blob/master/docs/media_features.md
**Note: JPEG encoding supports input format NV12/YUY2/UYVY/ABGR/Y8 and output format YUV400/YUV420/YUV422H_2Y/YUV444/RGB24.

PR #408, opened 2022-05-31
ffmpeg-qsv/ffmpeg-vaapi: enable hwmap testing

yefengxx:
add hwmap testing implement for ffmpeg-qsv and ffmpeg-vaapi

### DIFF
--- a/test/ffmpeg-vaapi/vpp/hwmap.py
+++ b/test/ffmpeg-vaapi/vpp/hwmap.py
@@ -1,0 +1,84 @@
+###
+### Copyright (C) 2022 Intel Corporation
+###
+### SPDX-License-Identifier: BSD-3-Clause
+###
+
+from ....lib import *
+from ....lib.ffmpeg.vaapi.util import *
+from ....lib.ffmpeg.vaapi.vpp import VppTest
+
+spec      = load_test_spec("vpp", "vaapi_hwmap")
+
+@slash.requires(*platform.have_caps("vpp", "scale"))
+@slash.requires(*have_ffmpeg_filter("scale_qsv"))
+class default(VppTest):
+  def before(self):
+    vars(self).update(
+      caps    = platform.get_caps("vpp", "scale"),
+    )
+    super(default, self).before()
+
+  def init(self, tspec, case, scale_width, scale_height, vpp_op, encoder):
+    vars(self).update(tspec[case].copy())
+    vars(self).update(
+      case          = case,
+      scale_height  = scale_height,
+      scale_width   = scale_width,
+      vpp_op        = vpp_op,
+      encoder       = encoder,
+    )
+
+  @slash.parametrize(*gen_vpp_hwmap_parameters(spec))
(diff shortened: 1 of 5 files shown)


yefengxx:
@uartie @FocusLuo  can you review this for me?

uartie:
Also, are there some technical docs about this feature?

test/ffmpeg-vaapi/vpp/hwmap.py
```diff
@@ -0,0 +1,84 @@
...
+###
+
+from ....lib import *
+from ....lib.ffmpeg.vaapi.util import *
+from ....lib.ffmpeg.vaapi.vpp import VppTest
+
+spec      = load_test_spec("vpp", "vaapi_hwmap")
+
+@slash.requires(*platform.have_caps("vpp", "scale"))
+@slash.requires(*have_ffmpeg_filter("scale_qsv"))
```
uartie, line 14: What if the test case isn't configured to use the "scale_qsv" vpp_op?

uartie:
I don't see all necessary checks for the requirements.  For example, what if scale_vulkan is not built/installed on target device?  It should skip the test case if that happens.

After a further quick look of this change, these feel more like transcode tests with vpp scaling.... just mixing different hw-based plugins to do the vpp and encode parts.  I wonder if it makes more sense to add this support to transcode tests, instead?

uartie:
In a bigger picture, this is more about mixing/interop between various decode, vpp, encode plugins.  The "hwmap" term is really only a ffmpeg concept and the means (syntax) to achieve interop between different hw-based codec elements in ffmpeg.

Furthermore, "raw" coding can also be categorized as a special codec for decode/encode in uncompressed/basic format (which we exploit a lot in this framework).  It just does not require extra "hwmap" syntax/concepts.  And the logic to handle this type of input/output is mostly encapsulated "under-the-hood". 

And what about gstreamer?  I'm sure we want to start testing mixed/interop elements here, too.  And it's likely more about caps/memory syntax in gstreamer rather than "hwmap" syntax.

Thus, I think "hwmap" (i.e. the logic to hook from one hw-element to another) should be totally encapsulated, too, since it can be defined differently between various middleware.

So, we really need to think about how we can **generalize** this concept of mixing elements/plugins without exposing middleware-specific terms into the test framework API's and especially into the higher-level user test-case configurations.

yefengxx:
> I don't see all necessary checks for the requirements. For example, what if scale_vulkan is not built/installed on target device? It should skip the test case if that happens.
> 
> After a further quick look of this change, these feel more like transcode tests with vpp scaling.... just mixing different hw-based plugins to do the vpp and encode parts. I wonder if it makes more sense to add this support to transcode tests, instead?

This test is for hwmap. The VPP scaling and encoding operations are just to verify whether the hwmap operation is successful. If we don't scale and encode, we won't know if hwmap is successful.

uartie:
Needs more thorough test design plan.  Close.

PR #415, opened 2022-07-04
add picture timing cases for ffmpeg-qsv/encode/hevc

fanli-hub:
add picture timing cases for ffmpeg-qsv/encode/hevc


fanli-hub:
@uartie @FocusLuo @Bin-QA, can you review this for me?

lib/parameters.py
```diff
@@ -279,6 +279,66 @@ def gen_avc_intref_lp_parameters(spec, profiles):
...
+        yield [
+          case, gop, bframes, bitrate, qp, variant["maxrate"],
+          profile, rcmode,
+        ]
+
+def gen_hevc_pict_lp_parameters(spec, profiles):
+  keys = ("case", "gop", "bframes", "bitrate", "qp", "maxrate", "profile", "rcmode")
+  params = gen_hevc_pict_lp_variants(spec, profiles)
+  return keys, params
+
```
uartie, line 341: This is exactly the same code as the non-lp variant.  Instead of duplication, how about we do:
```python
gen_hevc_pict_lp_parameters = gen_hevc_pict_parameters
```

uartie (changes requested):

lib/parameters.py
```diff
@@ -279,6 +279,66 @@ def gen_avc_intref_lp_parameters(spec, profiles):
...
+        ]
+
+def gen_hevc_pict_parameters(spec, profiles):
+  keys = ("case", "gop", "bframes", "bitrate", "qp", "maxrate", "profile", "rcmode")
+  params = gen_hevc_pict_variants(spec, profiles)
+  return keys, params
+
+def gen_hevc_pict_lp_variants(spec, profiles):
+  for case, params in spec.items():
+    for variant in copy.deepcopy(params.get("variants", dict()).get("pict", [])):
```
uartie, line 314: I guess you meant to use a different lookup variant... `pict_lp` not `pict`

lib/parameters.py
uartie: I guess you meant to use a different lookup variant... `pict_lp` not `pict` which would make them _slightly_ different.

uartie (changes requested):

lib/parameters.py
fanli-hub: > I guess you meant to use a different lookup variant... `pict_lp` not `pict` which would make them _slightly_ different.

Sorry for typing the wrong variant. I will modify this part.

fanli-hub:
> ...so why do we need special tests for this if we are not going to test the non-default option?

I have double-checked with Wenbin, he suggested that we should set " -pic_timing_sei 0" which means auto choose. Because msdk set default true, picture timing should work when we set " -pic_timing_sei 0".

lib/parameters.py
```diff
@@ -308,6 +338,7 @@ def gen_avc_max_frame_size_parameters(spec, profiles):
 gen_hevc_forced_idr_parameters = gen_avc_forced_idr_parameters
 gen_hevc_intref_lp_parameters = gen_avc_intref_lp_parameters
 gen_hevc_max_frame_size_parameters = gen_avc_max_frame_size_parameters
+gen_hevc_pict_lp_parameters = gen_hevc_pict_parameters
```
uartie, line 341: I know I suggested this alias at first, but later realized you have two variants that need handled, `pict` and `pict_lp`, so my first suggestion is wrong, hence the second suggestion/fix... sorry about the confusion from my end.

In your internal test config patches, you use `pict` and `pict_lp` variant keys to define test cases.  But this alias only parses the `pict` variant.  Thus, `pict_lp` is never used from the test config.  We should use separate variants and thus will need two separately defined param gen methods.  Perhaps the common code can be refactored and shared in one function, then just pass the lookup key depending on which variant it's for.

uartie (changes requested):

lib/parameters.py
fanli-hub: It's okay. I modified this part.

PR #419, opened 2022-07-13
lib/ffmpeg: decouple encoder/decoder from test class

uartie:
The encode/decode command line generator code is split into their own classes.  This allows them to be reused easier in different contexts/tests.  It also allows for simpler override for various properties based on plugin component differences.

cc: @wenbinc-Bin

### DIFF
--- a/lib/ffmpeg/d3d11/decoder.py
+++ b/lib/ffmpeg/d3d11/decoder.py
@@ -7,14 +7,11 @@
 import slash
 
 from ....lib.common import get_media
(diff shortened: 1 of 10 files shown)


uartie:
This also fixes a bug in ffmpeg encode `check_forced_idr` tests

uartie:
> > This also fixes a bug in ffmpeg encode `check_forced_idr` tests
> 
> I will split the bug fix into separate commit.

#421

PR #420, opened 2022-07-14
lib/gstreamer: use exe2os and filepath2os

tong1wu:

### DIFF
--- a/lib/gstreamer/vppbase.py
+++ b/lib/gstreamer/vppbase.py
@@ -6,7 +6,7 @@
 
 import slash
 
-from ...lib.common import timefn, get_media, call
+from ...lib.common import timefn, get_media, call, exe2os, filepath2os
 from ...lib.gstreamer.util import have_gst, have_gst_element
 from ...lib.metrics import md5
 from ...lib.mixin.vpp import VppMetricMixin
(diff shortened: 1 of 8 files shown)


uartie:
@tong1wu please fix test regressions.  HINT: don't use f-strings in `gen_xxx` functions.  The `gen_xxx` functions usually only return a "templated" string to be formatted later by caller.  That is, caller may use different values with the same returned template (e.g. r2r).

tong1wu:
> @tong1wu please fix test regressions. HINT: don't use f-strings in `gen_xxx` functions. The `gen_xxx` functions usually only return a "templated" string to be formatted later by caller. That is, caller may use different values with the same returned template (e.g. r2r).

Thanks for pointing out. I have replaced the f-strings with a new variable. Please help review again. Thanks.

uartie (changes requested):
Still alot of test failures.  Please don't use `self.decoded`, etc. in `gen_xxx` functions at all.  It might help to look at the ffmpeg classes to see how this is achieved for windows support.

lib/gstreamer/vppbase.py
```diff
@@ -50,6 +51,7 @@ def gen_input_opts(self):
     return opts
 
   def gen_output_opts(self):
+    self.osdecoded = filepath2os(self.decoded)
```
uartie, line 54: This should only be done in the `vpp()` function each time after `self.decoded` is assigned.

uartie: I know the current design can be confusing to know when and where certain variables should be accessed.  I'm hoping to clean this confusion up in some upcoming refactors soon.  In the meantime, we just have to be careful.

uartie:
I've changed this to "draft" since it's clear it has not been properly verified locally nor any help requested.

Once you're confident it works for you locally, you can switch it back to "ready for review".  If you need any help, I'd be happy to answer your questions... just ask.

tong1wu:
Tested all components. The can work now. Ready for review thanks.

uartie (approved):

PR #422, opened 2022-07-18
gst-va HEVC encode tests

uartie:


HeJunyan:
Basic OK for me. And I will analyse the issue if CI finds something.

https://github.com/intel/media-driver/issues/1390
is a known issue. For LP encoder, if reference > 3 and we get artifact pictures, it is this kind of problems.

PR #439, opened 2022-09-02
Add Max/min qp for I/P/B in ffmpeg-qsv/encode/avc

fanli-hub:
Add max/min qp for I/P/B frame in ffmpeg-qsv/encode/avc.


fanli-hub:
@FocusLuo @uartie, please help review.

FocusLuo:
@Bin-CI It seems like there are some checks failed, may I check with you whether this is caused by Lifan's patch?

uartie (approved):

PR #448, opened 2022-10-11
FFmpeg: use inline metrics

uartie:
Writing raw data to file, then reading into internal metric function adds unnecessary runtime overhead.

Instead, we can have FFmpeg calculate the MD5,SSIM, or PSNR concurrently.


uartie:
@Bin-CI can you verify this patch works on Windows, please?

FocusLuo (approved):
LGTM
Thank you very much Artie, hope this will decrease our testing execution time, and increase our testing efficiency

PR #450, opened 2022-10-18
lib/ffmpeg: escape and quote ssim/psnr filename for Windows support

uartie:
When a file path is used in ffmpeg filter option, the drive letter separator on Windows, ":", needs to be escaped so the ffmpeg command can parse it correctly. Otherwise, it will be mistaken as a filter option
separator.

### DIFF
--- a/lib/ffmpeg/decoderbase.py
+++ b/lib/ffmpeg/decoderbase.py
@@ -81,7 +81,7 @@ class Decoder(PropertyHandler, BaseFormatMapper):
         f" {self.ffdecoder} -r:v {fps} -i {self.ossource}"
         f" -f rawvideo -pix_fmt {self.format} -s:v {self.width}x{self.height}"
         f" -r:v {fps} -i {self.osreference}"
-        f" -lavfi '{self.scale_range},{mtype}=f={self.osstatsfile}:shortest=1'"
+        f" -lavfi \"{self.scale_range},{mtype}=f=\\'{self.osstatsfile}\\':shortest=1\""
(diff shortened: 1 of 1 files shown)


uartie:
This fixes a bug introduced in https://github.com/intel/vaapi-fits/pull/448

uartie:
> @Bin-CI can you test this patch to ensure it fixes Windows?
> 
> In particular, ffmpeg cases that use ssim or psnr metric are failing because of the erroneously parsed drive letter in the Windows filename. The mpeg, jpeg and vc1 decode cases use ssim. The encode cases use psnr. Please check these cases can work with this patch on Windows.

@Bin-CI also, when can we get pre-check bat tests on Windows?

uartie:
@Bin-CI can you experiment with the command line to see if you can find a syntax that works?

uartie:
> The filter path is special, how about change the filter to `-lavfi "null,ssim=f=\'c:/WINPATH/ddad3c8c-506c- 11ed-8641-d08e79100a6d_0_1_1.ssim\':shortest=1"` this format also passed on the verify BTW: `"\'\'"` can be parse on the test, `'\'\''` and `'\\'\\''` couldn't parse, it maybe refer windows part

Thanks!  Yes, that is better.  Patch updated.  It also works on Linux.

uartie:
Curious, is the escaped single quote necessary or can we just do `-lavfi "null,ssim=f='c:/WINPATH/ddad3c8c-506c- 11ed-8641-d08e79100a6d_0_1_1.ssim':shortest=1"`?

PR #452, opened 2022-10-25
ffmpeg/decode: WA for upstream regression 9990

uartie:
This is a temporary workaround for upstream regression reported in https://trac.ffmpeg.org/ticket/9990


uartie:
Superseded by #453

PR #454, opened 2022-10-26
ffmepg-qsv:Update TCBRC maxrate and bframes

Wanghangjie8:
If you don't set this parameter, h264 encode will automatically set to 0 but h265 won't,so need to set to 0 manually. maxrate needs to be equal to bitrate, but if maxrate is set to bitrate on ffmpeg, CBR will be used, so set maxrate=bitrate * 1.01.


FocusLuo (approved):
LGTM

PR #466, opened 2022-12-19
caps: remove rotation and transpose filter from dxva2 caps

tong1wu:
These filters are not supported by d3d9. Now remove them.


uartie (approved):
LGTM

PR #468, opened 2022-12-25
[Gst-va]Add encode and transcode in gst va

Wanghangjie8:


Wanghangjie8:
@Bin-QA can you review this for me?

PR #472, opened 2023-01-06
lib/parameters : update lib/parameters

Wanghangjie8:


uartie (changes requested):
Do we have need for non-lowpower av1?

Wanghangjie8:
Hi @uartie ,Ffmpeg-sw/encode/av1.py does not need lp, so I added this definition in vaapi-fits.

uartie:
check failures caused by upstream regression in ffmpeg-vaapi commit fbdba9a1a69fe4df413d9e9df1b11db522946e75

PR #473, opened 2023-01-18
Update ffmpeg av1 doesn't support sw

wenqingx:
Signed-off-by: Wang, WenqingX <wenqingx.wang@intel.com>

### DIFF
--- a/lib/ffmpeg/transcoderbase.py
+++ b/lib/ffmpeg/transcoderbase.py
@@ -205,7 +205,15 @@ class BaseTranscoderTest(slash.Test,BaseFormatMapper):
           "{}_{}_{}.yuv".format(self.case, n, channel))
         osyuv = filepath2os(yuv)
         vppscale = self.get_vpp_scale(self.width, self.height, "sw")
-        iopts = "-i {}"
+        iopts = ""
+        if output["codec"] in ["av1"]:
+          # WA: no SW decoder for AV1 in ffmpeg, currently
+          ###
+          # TODO: setup hwaccel stuff
+          ###
+          decoder = self.get_decoder(output["codec"], "hw")
+          iopts += f"-c:v {decoder}"
+        iopts += " -i {}"
         oopts = "-vf '{}' -pix_fmt yuv420p -f rawvideo -vframes {} -y {}"
         self.call_ffmpeg(
           iopts.format(osencoded), oopts.format(vppscale, self.frames, osyuv))
--- a/lib/ffmpeg/vaapi/transcoder.py
+++ b/lib/ffmpeg/vaapi/transcoder.py
@@ -36,7 +36,7 @@ class TranscoderTest(BaseTranscoderTest):
         hw = (platform.get_caps("decode", "vc1"), have_ffmpeg_decoder("vc1"), "vc1"),
       ),
       "av1" : dict(
-        hw = (platform.get_caps("decode", "av1_8"), have_ffmpeg_decoder("av1"), "av1"),
+        hw = (platform.get_caps("decode", "av1_8"), have_ffmpeg_decoder("av1_qsv"), "av1_qsv"),
       ),
     },
     encode = {


wenqingx:
Hi @uartie @Bin-CI ,ffmpeg does not support av1 sw now, so an error will be reported during the check. I try to modify it this way and use hw. Would you have any better suggestions?thanks

lib/ffmpeg/transcoderbase.py
```diff
@@ -205,7 +205,12 @@ def transcode(self):
           "{}_{}_{}.yuv".format(self.case, n, channel))
         osyuv = filepath2os(yuv)
         vppscale = self.get_vpp_scale(self.width, self.height, "sw")
+        codec = output["codec"]
+        mode = output["mode"]
+        encoder = self.get_encoder(codec, mode)
```
uartie, line 210: This should be `self.get_decoder(codec, mode)`.

lib/ffmpeg/transcoderbase.py
```diff
@@ -205,7 +205,12 @@ def transcode(self):
...
         iopts = "-i {}"
+        if ("av1_qsv") == encoder:
+          iopts = "-c:v av1_qsv -i {}"
```
uartie, line 213: Maybe it would be better if we refactor some logic so we can specialize this in lib/ffmpeg/qsv/transcoder.py instead.

uartie (changes requested):

lib/ffmpeg/transcoderbase.py
uartie, line 213: This needs to be more general so that it can work with other hwaccel plugins (qsv, vaapi, etc).  Basically, we'll need to setup the hwaccel like we do in `gen_input_opts`.  Also, this is the only special case at the moment, so it should be something like:
```python
iopts = ""
if output["codec"] in ["av1"]:
  # WA: no SW decoder for AV1 in ffmpeg, currently
  ###
  # TODO: setup hwaccel stuff
  ###
  decoder = self.get_decoder(output["codec"], "hw")
  iopts += f"-c:v {decoder}"

iopts += " -i {}"
```

uartie: Option 2: It might be possible to use the new Decoder classes in lib/ffmpeg/qsv/decoder.py, lib/ffmpeg/vaapi/decoder.py, etc. with some minor changes.

wenqingx: thanks for you advice, ffmpeg-qsv is passed.But i need to modify /lib/ffmpeg/vaapi/transcoder.py av1 codec when  running ffmpeg-vaapi. If not the {decoder} is av1_qsv,It is still not supported during verification.

lib/ffmpeg/vaapi/transcoder.py
```diff
@@ -36,7 +36,7 @@ class TranscoderTest(BaseTranscoderTest):
         hw = (platform.get_caps("decode", "vc1"), have_ffmpeg_decoder("vc1"), "vc1"),
       ),
       "av1" : dict(
-        hw = (platform.get_caps("decode", "av1_8"), have_ffmpeg_decoder("av1"), "av1"),
+        hw = (platform.get_caps("decode", "av1_8"), have_ffmpeg_decoder("av1_qsv"), "av1_qsv"),
```
wenqingx, line 39: @uartie ,can I modify it directly?

uartie: Right.  Please don't use qsv plugins in vaapi context.  We should not assume qsv is available when running vaapi cases.

wenqingx: emmm,but when I run vaapi, iopts will get "ffmpeg - v verbose - c: v av1 ...", It is still not supported during verification.

lib/ffmpeg/transcoderbase.py
uartie: I'd prefer not to introduce too many special-case hacks.  Please create an internal JIRA and assign to me... this corner-case was not anticipated in the original code and may require more refactoring than expected.  I have some ideas and may be able to help by providing a solution or examples.

lib/ffmpeg/vaapi/transcoder.py
uartie: Because you need to setup the vaapi hwaccel parameters (i.e. see the TODO in my review)

uartie: OOP principles are necessary.

lib/ffmpeg/transcoderbase.py
wenqingx: OK, I created a interal JIRA and assign to you ,thanks for your help.

uartie:
superseded by #476

PR #478, opened 2023-02-01
lib/caps : remove VC1 decode test case input

jian2x:
Signed-off-by: jian13x <jian2x.zhou@intel.com>


FocusLuo (approved):
LGTM

FocusLuo:
@uartie @Bin-CI 
there is almost no any user for VC1 decode, and ffmpeg/gstreamer dev will not pay any effort for it, more ever start from DG2/ATSM, vc1 decode are not supported by iHD driver. 
Based on gst forum sync result and team internal suggestion, how about let us to remove all VC1 decode test cases?

uartie:
I agree with @Bin-CI, these platforms are still capable of VC1... so don't remove the caps.  Remove the user-defined cases instead.

PR #479, opened 2023-02-09
[ffmpeg-vaapi] Delete the check of IPB mode

Wanghangjie8:


Wanghangjie8:
When we delete the check of IPB mode, the passrate of vaapi-fits-quality will be improved.

Wanghangjie8:
@Bin-CI @uartie can you review this for me?

jian2x:
I remember that uartie mentioned a PR about IPB mode before, but the patch of uartie seems to have no effect on the DG2 platform

uartie:
Just because the encode command can succeed, it does not mean it did the right thing.  Hence, these extra sanity checks.

PR #485, opened 2023-03-03
[gstreamer]enable av1 8bit vdenc transcode for gst 

wenqingx:


uartie:
Do you have a PR for bat/sanity with some new cases?

wenqingx:
> Do you have a PR for bat/sanity with some new cases?

I have previously submitted some test cases, but gst was skipped at the time. If this patch is merged, need to add refs for those

PR #486, opened 2023-03-15
[ffmpeg/gstreamer] Skip the bitrate_check of the specified case

Wanghangjie8:


Wanghangjie8:
Hi @uartie , can you review this for me?

lib/ffmpeg/encoderbase.py
```diff
@@ -33,6 +33,7 @@ class Encoder(PropertyHandler, BaseFormatMapper):
   encoded       = property(lambda s: s._encoded)
   osencoded     = property(lambda s: filepath2os(s.encoded))
   encoded_ext   = property(lambda s: s.props["encoded_ext"])
+  skip_checks   = property(lambda s: s.props["skip_checks"])
```
uartie, line 36: This needs to be an optional property.  If not specified, it should return an empty list.  Otherwise, we are going to break other test cases (as seen in CI pre-check results).

lib/gstreamer/encoderbase.py
```diff
@@ -34,6 +34,7 @@ class Encoder(PropertyHandler):
   encoded       = property(lambda s: s._encoded)
   osencoded     = property(lambda s: filepath2os(s.encoded))
   encoded_ext   = property(lambda s: s.props["encoded_ext"])
+  skip_checks   = property(lambda s: s.props["skip_checks"])
```
uartie, line 37: needs to be optional property

uartie (changes requested):

lib/ffmpeg/encoderbase.py
uartie, line 36: This should not be a property of the `Encoder` class.  In fact, you are not even using this property from the `EncoderTest` class.  If you were, then in that class it would be `self.encoder.skip_checks`, not `self.skip_checks`.   Nonetheless, it does not belong here.

Please remove this property from the Encoder classes.

lib/ffmpeg/encoderbase.py
```diff
@@ -239,14 +240,14 @@ def check_bitrate(self):
       size_encoded = encsize,
       bitrate_actual = "{:-.2f}".format(bitrate_actual))
 
-    if "cbr" == self.rcmode:
+    if "cbr" == self.rcmode and "bitrate_gap" not in self.skip_checks:
```
uartie, line 243: Use `vars(self).get("skip_checks", [])`, instead, here and in the other places... because this is an optional parameter.

uartie (changes requested):

lib/ffmpeg/encoderbase.py
Wanghangjie8: Thank you for your review, I have made corresponding modifications to it

uartie (changes requested):

lib/ffmpeg/encoderbase.py
```diff
@@ -239,14 +239,14 @@ def check_bitrate(self):
       size_encoded = encsize,
       bitrate_actual = "{:-.2f}".format(bitrate_actual))
 
-    if "cbr" == self.rcmode:
+    if "cbr" == self.rcmode and vars(self).get("skip_check", []):
```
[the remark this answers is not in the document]
Wanghangjie8: Thank you for your review, I have made corresponding modifications to it

Wanghangjie8:
Hi @uartie  , can you review this for me?

lib/ffmpeg/encoderbase.py
```diff
@@ -238,15 +238,17 @@ def check_bitrate(self):
     get_media()._set_test_details(
       size_encoded = encsize,
       bitrate_actual = "{:-.2f}".format(bitrate_actual))
+    if "bitrate_gap" in vars(self).get("skip_check", []):
+      return
 
-    if "cbr" == self.rcmode:
+    if "cbr" == self.rcmode and vars(self).get("skip_check", []):
```
uartie, line 244: This change is not needed in the conditionals anymore.  You've already checked it above.

Wanghangjie8: But, after I tried to delete this change, the case still has a bitrate_gap issues

lib/ffmpeg/encoderbase.py
```diff
@@ -238,15 +238,17 @@ def check_bitrate(self):
     get_media()._set_test_details(
       size_encoded = encsize,
       bitrate_actual = "{:-.2f}".format(bitrate_actual))
+    if "bitrate_gap" in vars(self).get("skip_check", []):
```
uartie, line 241: `skip_checks` (plural)

lib/ffmpeg/encoderbase.py
uartie: I can't see what you're actually doing that leads you to that conclusion.  But you're doing something wrong because if I fix it locally, it works for me.

lib/ffmpeg/encoderbase.py
```diff
@@ -238,15 +238,17 @@ def check_bitrate(self):
     get_media()._set_test_details(
       size_encoded = encsize,
       bitrate_actual = "{:-.2f}".format(bitrate_actual))
+    if "bitrate_gap" in vars(self).get("skip_checks", []):
+      return
```
uartie, line 242: We should add a warning message here, too.  (slash.logger.warn...)

lib/ffmpeg/encoderbase.py
```diff
@@ -7,6 +7,7 @@
 import os
 import re
 import slash
+import logging
```
uartie, line 10: we don't use the logging module... use slash.logger...

uartie (changes requested):

uartie:
superseded by #490

PR #488, opened 2023-03-23
lib/ffmpeg/vaapi/encoder.py #do not check ipb mode when use default s…

jian2x:
Signed-off-by: Zhoujian [[jian2x.zhou@intel.com]

### DIFF
--- a/lib/ffmpeg/vaapi/encoder.py
+++ b/lib/ffmpeg/vaapi/encoder.py
@@ -60,10 +60,11 @@ class EncoderTest(BaseEncoderTest):
 
   def check_output(self):
     # profile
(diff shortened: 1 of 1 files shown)


jian2x:
@uartie  please help review whether it is reasonable，thanks

uartie (changes requested):
It is not clear what you are trying to solve.  Please explain.  And fix your commit message!

lib/ffmpeg/vaapi/encoder.py
uartie, line 1: It looks like you are trying to workaround special use-case's (defaults) imposed by our internal quality suite.  It would be better to just override the `check_output` method in the derived test classes in the quality suite.

PR #490, opened 2023-04-04
encode: add option to skip bitrate gap check

uartie:
Some user-defined cases may not have enough frames to effectively achieve the desired bitrate.

For such cases, this enables the user to add
skip_checks = ['bitrate_gap'] to the case definition.

This supersedes #486


FocusLuo (approved):
LGTM

PR #491, opened 2023-04-06
Add gstreamer HEVC VME 12bit 420 and 422 tests

uartie:
TGL, ADL, DG1, and SG1 support HEVC VME 12bit 4:2:0 and 4:2:2 encoding.

gst-va supports 4:2:0 and 4:2:2 12bit profiles.
gst-vaapi and gst-msdk only support 4:2:0 12bit profiles.

### DIFF
--- a/test/gst-msdk/encode/12bit/hevc.py
+++ b/test/gst-msdk/encode/12bit/hevc.py
@@ -1,0 +1,100 @@
+###
+### Copyright (C) 2023 Intel Corporation
+###
+### SPDX-License-Identifier: BSD-3-Clause
+###
+
+from .....lib import *
+from .....lib.gstreamer.msdk.util import *
+from .....lib.gstreamer.msdk.encoder import EncoderTest
+
+spec = load_test_spec("hevc", "encode", "12bit")
+
+@slash.requires(*have_gst_element("msdkh265dec"))
+@slash.requires(*have_gst_element("msdkh265enc"))
+class HEVC12EncoderBaseTest(EncoderTest):
+  def before(self):
+    super().before()
+    vars(self).update(
+      codec         = "hevc-12",
+      gstdecoder    = "msdkh265dec",
+      gstencoder    = "msdkh265enc",
+      gstmediatype  = "video/x-h265",
+      gstparser     = "h265parse",
+    )
+
+  def get_file_ext(self):
+    return "h265"
+
+
+@slash.requires(*platform.have_caps("encode", "hevc_12"))
+class HEVC12EncoderTest(HEVC12EncoderBaseTest):
+  def before(self):
+    super().before()
+    vars(self).update(
+      caps      = platform.get_caps("encode", "hevc_12"),
+      lowpower  = False,
+    )
+
+class cqp(HEVC12EncoderTest):
+  def init(self, tspec, case, gop, slices, bframes, qp, quality, profile):
+    vars(self).update(tspec[case].copy())
+    vars(self).update(
+      bframes   = bframes,
+      case      = case,
+      gop       = gop,
+      profile   = profile,
+      qp        = qp,
+      quality   = quality,
+      rcmode    = "cqp",
+      slices    = slices,
+    )
+
+  @slash.parametrize(*gen_hevc_cqp_parameters(spec, ['main12']))
+  def test(self, case, gop, slices, bframes, qp, quality, profile):
+    self.init(spec, case, gop, slices, bframes, qp, quality, profile)
+    self.encode()
+
+class cbr(HEVC12EncoderTest):
+  def init(self, tspec, case, gop, slices, bframes, bitrate, fps, profile):
+    vars(self).update(tspec[case].copy())
+    vars(self).update(
+      bframes   = bframes,
+      bitrate   = bitrate,
+      case      = case,
+      fps       = fps,
+      gop       = gop,
+      maxrate   = bitrate,
+      minrate   = bitrate,
+      profile   = profile,
+      rcmode    = "cbr",
+      slices    = slices,
+    )
+
+  @slash.parametrize(*gen_hevc_cbr_parameters(spec, ['main12']))
+  def test(self, case, gop, slices, bframes, bitrate, fps, profile):
+    self.init(spec, case, gop, slices, bframes, bitrate, fps, profile)
+    self.encode()
+
+class vbr(HEVC12EncoderTest):
+  def init(self, tspec, case, gop, slices, bframes, bitrate, fps, quality, refs, profile):
+    vars(self).update(tspec[case].copy())
+    vars(self).update(
+      bframes   = bframes,
+      bitrate   = bitrate,
+      case      = case,
+      fps       = fps,
+      gop       = gop,
+      maxrate   = bitrate * 2,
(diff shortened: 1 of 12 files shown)


test/gst-vaapi/encode/12bit/hevc.py
```diff
@@ -0,0 +1,99 @@
...
+  def init(self, tspec, case, gop, slices, bframes, bitrate, fps, quality, refs, profile):
+    vars(self).update(tspec[case].copy())
+    vars(self).update(
+      bframes   = bframes,
+      bitrate   = bitrate,
+      case      = case,
+      fps       = fps,
+      gop       = gop,
+      maxrate   = int(bitrate / 0.7),
+      minrate   = bitrate,
```
xhaihao, line 88: If so, minrate will be greater than maxrate, right ? 

I guess this was copied & pasted from 10bit, but according to the comment in commit 4767de24faa96ce26319ed8c4e0a64dd7d593bd4,  Both maxrate and minrate were set incorrectly for 10bit.
+      ## gst-vaapi sets max-bitrate = bitrate and min-bitrate = bitrate * 0.70
+      maxrate      = int(bitrate / 0.7),
+      minrate      = bitrate,

test/gst-va/encode/12bit/hevc.py
```diff
@@ -0,0 +1,99 @@
...
+  def init(self, tspec, case, gop, slices, bframes, bitrate, fps, quality, refs, profile):
+    vars(self).update(tspec[case].copy())
+    vars(self).update(
+      bframes   = bframes,
+      bitrate   = bitrate,
+      case      = case,
+      fps       = fps,
+      gop       = gop,
+      maxrate   = int(bitrate / 0.7),
+      minrate   = bitrate,
```
xhaihao, line 88: minrate will be greater than maxrate, see another comment.

test/gst-msdk/encode/12bit/hevc.py
```diff
@@ -0,0 +1,100 @@
...
+  def init(self, tspec, case, gop, slices, bframes, bitrate, fps, quality, refs, profile):
+    vars(self).update(tspec[case].copy())
+    vars(self).update(
+      bframes   = bframes,
+      bitrate   = bitrate,
+      case      = case,
+      fps       = fps,
+      gop       = gop,
+      maxrate   = bitrate * 2,
+      minrate   = bitrate,
```
xhaihao, line 89: Do we need to set maxrate and minrate in the same way for gst-va, gst-vaapi and gst-msdk ?

test/gst-vaapi/encode/12bit/hevc.py
uartie: Yes, I understand how this could alarm you or be a little confusing ;)

From the vaapi-fits perspective, we want "bitrate" to be normalized from the user test case config and when internally validating the bitrate gap.  That is, when user defines a vaapi-fits test case with a specific bitrate, the goal is to have the libva driver receive the same values (bits_per_second and target_percentage), regardless of which plugins and middleware is handling it.

Unfortunately, each plugin/middleware has different properties and/or applies different rules/functions to the intermediate values before sending to the libva driver.  Originally, gst-vaapi hard-coded the target-percentage to 70%... but later we added this property to gst-vaapi so we could configure it.  However, we kept it at 70% in the test framework to avoid needing to re-baseline all of our PSNR refs.  Either way, here is how the values are propagated from vaapi-fits to the gst-vaapi/gst-va plugins to achieve bitrate normalization (gst-msdk omitted here as an exercise to reader, if interested).

For gst-vaapi:
```
 test.bitrate = user.bitrate
 test.maxrate = test.bitrate / 0.7
 test.minrate = test.bitrate
 gst-vaapi.target_percentage = test.minrate / test.maxrate * 100
 gst-vaapi.bitrate = test.maxrate

 # inside gstreamer vaapi encoder
 libva.bits_per_second = gst-vaapi.bitrate * 1000
 libva.target_percentage = gst-vaapi.target_percentage
```

For gst-va:
```
 test.bitrate = user.bitrate
 test.maxrate = test.bitrate / 0.7
 test.minrate = test.bitrate
 gst-va.target_percentage = test.minrate / test.maxrate * 100
 gst-va.bitrate = test.minrate

 # inside gstreamer va encoder
 libva.bits_per_second = (gst-va.bitrate * 100 / gst-va.target_percentage) * 1000
 libva.target_percentage = gst-va.target_percentage
```

So, for example, if user sets `bitrate = 500` (user.bitrate) at test config, then `libva.target_percentage = 70` and `libva.bits_per_second = 714` for both gst-vaapi and gst-va plugins.  And similarly, should be for gst-msdk too.

test/gst-va/encode/12bit/hevc.py
uartie: see reply in https://github.com/intel/vaapi-fits/pull/491#discussion_r1160829998

test/gst-msdk/encode/12bit/hevc.py
uartie: see reply in https://github.com/intel/vaapi-fits/pull/491#discussion_r1160829998

uartie: However, I think gst-msdk will end up with 50% target-percentage.... which is in line with our ffmpeg default tests settings.  The 70% for gst-vaapi/gst-va was a left-over side-effect of it being hard-coded previously.

test/gst-vaapi/encode/12bit/hevc.py
uartie: Short answer:
No.  The `maxrate` will be greater than `minrate`.  Note the division by a fraction makes the `maxrate` greater.  For example, bitrate=500, **maxrate** = 500 / 0.7 = **714**, **minrate** = bitrate = **500**.

test/gst-msdk/encode/12bit/hevc.py
uartie: IIUC, the gst-va refs were copied from gst-vaapi.  In the future, we should make them all use 50% target but will require new PSNR refs.

uartie: Yes, good point.

That was the original idea we wanted.  Unfortunately, when we wrote the other tests, gst-vaapi used to hard-code target-percentage to 70%.  So it was not possible to make minrate 50% of maxrate in gst-vaapi.  But now gst-vaapi has the property to allow this.  We kept it at 70% for existing tests to avoid rebasing PSNR refs.  And when we copied the old test to make new ones, we should have changed the new ones to be consistent, but it was forgotten :(.

Pretty sure we use 50% in ffmpeg tests, too.

Since these are new tests, I will make them the same.  Thanks.

test/gst-vaapi/encode/12bit/hevc.py
xhaihao: I am sorry I made a stupid error. I took it as 'maxrate=int(bitrate * 0.7).

FocusLuo (approved):
LGTM for this patch for Gen12- platforms. In case for next supports features,  two another comments:
1, Gen12.7+ will only have LP encode path
2, Gen12.y+ will support 16k max resolutions

PR #494, opened 2023-04-26
vpp/dn: normalize vaapi vpp dn levels

uartie:
Normalize test framework DN levels for ffmpeg-vaapi and gst-vaapi so the specified user config vpp level ultimately maps to the same value that middleware
sends to the libva driver.

The mfx middleware plugins are already specified in the same range as the test framework.

This will enable DN refs to be shared across components.

cc: @tong1wu


FocusLuo (approved):
LGTM

tong1wu (approved):

PR #495, opened 2023-05-05
gst/decode: link avvideocompare to fakevideosink

uartie:
The avvideocompare does not work properly if the data is not aligned.  To force alignment, downstream needs to advertise video meta.  Thus, link avvideocompare to fakevideosink to inject video meta into the pipeline.


uartie:
TODO: need to enable avvideocompare build in pre-check to verify

uartie:
> TODO: need to enable avvideocompare build in pre-check to verify

Enabled. KBL regressed.  Added qos=false to fakevideosink to fix it.

PR #500, opened 2023-05-10
test/gst-va: add av1e 10bit vdenc tests

uartie:


FocusLuo (approved):
LGTM

FocusLuo:
> LGTM, BTW: gst-va enc/av1 without non-lp support?
Right,  av1 encode is using fixing function hardware

PR #501, opened 2023-05-10
gst/va/encoder: handle tilerows/cols

uartie:
The tilerows and tilecols properties are specified in AV1e tests but encoder base forgot to use them.

Add tilerows and tilecols properties to the gst command-line.


FocusLuo (approved):
LGTM

PR #506, opened 2023-05-18
ffmpeg: no dump for inline psnr/ssim

uartie:
When doing inline psnr/ssim, the raw conversion and output file dump is unnecessary and only adds runtime overhead.

Remove the conversion and file dump to improve inline psnr/ssim speed.


FocusLuo (approved):
LGTM

PR #508, opened 2023-05-31
add avc low power mode bframes support for ffmpeg/gstreamer

FocusLuo:
1, lib parameter core: support avc lp mode bframes;  and let hevc lp mode to gen itself  lp cqp/cbr/vbr parameters
2, ffmpeg-vaapi avc lp cqp/cbr/vbr and r2r support bframes
3, ffmpeg-qsv avc lp cqp/cbr/vbr and r2r support bframes
4, gst-vaapi avc lp cqp/cbr/vbr and r2r support bframes
5, gst-va avc lp cqp/cbr/vbr and r2r support bframes
6, gst-msdk avc lp cqp/cbr/vbr and r2r support bframes
7, [Caps] define there will no bframes on pre-gen12.7 platforms
8, skip bframes check for both ffmpeg and gstreamer on their encoderbase.py


FocusLuo:
@uartie Can you help to give a review for this?

uartie (changes requested):
Please fix your CI failures, first

uartie:
> @uartie Can you help to give a review for this?

Can you provide a problem statement about what/why we need these changes?  Why existing tests can't cover the use-case or why we can't just define a new test?

FocusLuo:
> > @uartie Can you help to give a review for this?
> 
> Can you provide a problem statement about what/why we need these changes? Why existing tests can't cover the use-case or why we can't just define a new test? Also, are we sure these changes don't affect/break/change other tests?

@uartie currently 

1. There was no avc b frames support in low power mode, so there are no test cases cover before (with bframes=0 setup)
2.  avc LP b frames will only support Gen12.7+ platforms (that is why bframes = False are added in caps for pre-Gen12.7 platforms define) 
3. now we add some test cases with bframes>0 in sanity/full/bat
4. Right, they will not impact current tests.

test/ffmpeg-qsv/encode/avc.py
```diff
@@ -82,13 +83,13 @@ def init(self, tspec, case, gop, slices, qp, quality, profile):
     )
 
   @slash.parametrize(*gen_avc_cqp_lp_parameters(spec, ['high', 'main']))
-  def test(self, case, gop, slices, qp, quality, profile):
```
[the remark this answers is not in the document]
FocusLuo: @uartie I understand your concern. Yes, It will lost history data. Actually I also think about this before. 
One thing need to let you do compare, because existing test cases only do test for I, IP frame mode for LP AVC CQP/CBR/VBR with high/main profile and a lot of resolution, slices (...ect, parameter), From user side, IPB mode is more usage than I/IP mode, so IPB will also need to cover these testing parameter.
That means if to submit new test cases for test_bframes, there will have another new test functions name of gen_avc_cqp_lp_bframes_parameters/gen_avc_cbr_lp_bframes_parameters/gen_avc_vbr_lp_bframes_parameters..., and will have new class with names of cqp_lp_bframes, cbr_lp_bframes, vbr_lp_bframes almost same test parameters with cqp_lp/cbr_lp/vbr_lp in each ffmpeg/gstramer encode/avc.py files. This is why I submitted vaapi-fits patches with all full/sanity/bat AVC LP encode test cases reference rebased. Currently, existing AVC LP encode test case results are good, maybe we can use this chance to refined the AVC LP test reference.
How do you think?

lib/ffmpeg/encoderbase.py
```diff
@@ -184,6 +184,11 @@ def validate_caps(self):
         format_value(
           "{platform}.{driver}.slice > 1 unsupported in this mode", **vars(self)))
 
+    if vars(self).get("bframes", 2) > 0 and not self.caps.get("bframes", True):
```
uartie, line 187: Why `2` as default?  If test does not set bframes, then should we assume 0?

lib/parameters.py
```diff
@@ -454,9 +503,9 @@ def gen_hevc_pict_lp_parameters(spec, profiles):
 gen_hevc_cbr_parameters = gen_avc_cbr_parameters
 gen_hevc_vbr_parameters = gen_avc_vbr_parameters
 gen_hevc_roi_parameters = gen_avc_roi_parameters
-gen_hevc_cqp_lp_parameters = gen_avc_cqp_lp_parameters
-gen_hevc_cbr_lp_parameters = gen_avc_cbr_lp_parameters
-gen_hevc_vbr_lp_parameters = gen_avc_vbr_lp_parameters
+#gen_hevc_cqp_lp_parameters = gen_avc_cqp_lp_parameters
+#gen_hevc_cbr_lp_parameters = gen_avc_cbr_lp_parameters
+#gen_hevc_vbr_lp_parameters = gen_avc_vbr_lp_parameters
```
uartie, line 508: Just delete these.

FocusLuo: deleted and update

lib/ffmpeg/encoderbase.py
FocusLuo: updated

uartie:
Please rebase.

FocusLuo:
> Please rebase.

Rebased, thanks

PR #510, opened 2023-06-05
add per pixel blending tests

xhaihao:


uartie (changes requested):

test/ffmpeg-qsv/vpp/overlay.py
```diff
@@ -0,0 +1,31 @@
...
+### Copyright (C) 2023 Intel Corporation
+###
+### SPDX-License-Identifier: BSD-3-Clause
+###
+
+from ....lib import *
+from ....lib.ffmpeg.qsv.util import *
+from ....lib.ffmpeg.qsv.vpp import VppTest
+
+spec = load_test_spec("vpp", "overlay", "rgba")
```
uartie, line 11: "rgba" seems unnecessary in this spec key... nothing about the test implementation explicitly requires it.  It is up to user test case config to supply their own `source1` which is not constrained to any particular format in the test even if other formats may not work or make sense.

uartie (changes requested):

uartie (changes requested):

uartie:
I think we should also allow x,y and w,h user test params... they can be optional

test/ffmpeg-qsv/vpp/overlay.py
xhaihao: Maybe this will be needed in future, e.g. for a rgb (rgb0) case, the output format from ffmpeg sw decoder is gbrp format, we have to convert it to rgb0

xhaihao:
> I think we should also allow x,y and w,h user test params... they can be optional

These are covered in composite cases. The intention of adding these cases is to make sure per pixel alpha blending can work. But I agree we may add more cases to cover x/y, w/h in future.

test/ffmpeg-qsv/vpp/overlay.py
xhaihao: I will remove it in this spec. We may add it back if it is needed in future.

lib/ffmpeg/vaapi/vpp.py
```diff
@@ -36,6 +36,9 @@ def gen_vpp_opts(self):
       vpfilter.append(f"{finputs}{self.stack}_vaapi=inputs={self.inputs}")
       if self.stack in ["xstack"]:
         vpfilter[-1] += f":grid={self.cols}x{self.rows}:grid_tile_size={self.tilew}x{self.tileh}"
+    elif self.vpp_op in ["overlay"]:
+      vpfilter.append("[1:v]hwupload[v1];")
+      vpfilter[-1] += "[0:v][v1]overlay_vaapi=alpha={ra}"
```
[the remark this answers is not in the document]
xhaihao: updated, thx

test/ffmpeg-qsv/vpp/overlay.py
```diff
@@ -0,0 +1,31 @@
...
+    )
+    super().before()
+
+  @slash.parametrize(*gen_vpp_overlay_parameters(spec))
+  def test(self, case, alpha):
+    vars(self).update(spec[case].copy())
+    vars(self).update(
+      case    = case,
+      alpha   = alpha,
+      ra      = int(alpha * 255)
```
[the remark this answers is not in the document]
xhaihao: updated, thx

test/ffmpeg-qsv/vpp/overlay.py
xhaihao: removed, thx

test/ffmpeg-vaapi/vpp/overlay.py
```diff
@@ -0,0 +1,31 @@
...
+    )
+    super().before()
+
+  @slash.parametrize(*gen_vpp_overlay_parameters(spec))
+  def test(self, case, alpha):
+    vars(self).update(spec[case].copy())
+    vars(self).update(
+      case    = case,
+      alpha   = alpha,
+      ra      = alpha,
```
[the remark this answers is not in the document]
xhaihao: updated, thx

PR #515, opened 2023-06-08
gst-msdk: Add surface sharing cases between va and msdk

mengker33:
These cases are about implicit va surface sharing between va and msdk, i.e. "vah264dec ! msdkh264enc ! h264parse", in which vah264dec and msdkh264enc sharing va surface via negotiated va memory caps.

### DIFF
--- a/lib/gstreamer/msdk/transcoder.py
+++ b/lib/gstreamer/msdk/transcoder.py
@@ -21,10 +21,12 @@ class TranscoderTest(BaseTranscoderTest):
       "avc" : dict(
         sw = (dict(maxres = (16384, 16384)), have_gst_element("openh264dec"), "h264parse ! openh264dec ! videoconvert"),
         hw = (platform.get_caps("decode", "avc"), have_gst_element("msdkh264dec"), "h264parse ! msdkh264dec"),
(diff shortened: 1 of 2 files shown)


lib/gstreamer/msdk/transcoder.py
```diff
@@ -15,16 +15,19 @@
 
 @slash.requires(*have_gst_element("msdk"))
 @slash.requires(using_compatible_driver)
+@slash.requires(*have_gst_element("va"))
```
uartie, line 18: We should not have this here.  If user disables gst-va plugin build and just enables gst-msdk plugin build, then they cannot test gst-msdk cases only because this `slash.requires` would fail and cause ALL transcode cases to skip.

The `have_gst_element("va...")` in the `requirements` dict (below) is enough to have va_hw cases automatically skipped in that scenario.

uartie (changes requested):

lib/gstreamer/msdk/transcoder.py
mengker33: Understood. I removed this line, thanks

uartie (approved):
LGTM. Thanks!

PR #516, opened 2023-06-08
[Gst-va] add Vpp crop support

wenqingx:


uartie (changes requested):
This is incomplete.  The crop parameters also need to be mapped to the gst-va command-line in `lib/gstreamer/va/vpp.py:gen_vpp_opts(...)`.  See `lib/gstreamer/vaapi/vpp.py` for an example.

test/gst-va/vpp/crop.py
```diff
@@ -0,0 +1,35 @@
...
+
+@slash.requires(*platform.have_caps("vpp", "crop"))
+class default(VppTest):
+  def before(self):
+    vars(self).update(
+      caps    = platform.get_caps("vpp", "crop"),
+      metric  = dict(type = "md5"),
+      vpp_op  = "crop",
+    )
+    super(default, self).before()
```
uartie, line 21: `super().before()` is better

uartie:
> This is incomplete. The crop parameters also need to be mapped to the gst-va command-line in `lib/gstreamer/va/vpp.py:gen_vpp_opts(...)`. See `lib/gstreamer/vaapi/vpp.py` for an example.

Ah, actually the vapostproc plugin uses crop meta for crop support.  Thus, we need to insert a videocrop before the vapostproc.  This can be done inside the `lib/gstreamer/va/vpp.py:gen_output_opts(...)` function override.

uartie:
https://gstreamer.freedesktop.org/documentation/va/vapostproc.html?gi-language=c#cropping-example

uartie:
> > This is incomplete. The crop parameters also need to be mapped to the gst-va command-line in `lib/gstreamer/va/vpp.py:gen_vpp_opts(...)`. See `lib/gstreamer/vaapi/vpp.py` for an example.
> 
> Ah, actually the vapostproc plugin uses crop meta for crop support. Thus, we need to insert a videocrop before the vapostproc. This can be done inside the `lib/gstreamer/va/vpp.py:gen_output_opts(...)` function override.

Also, we should set `disable-passthrough=true` to force vapostproc to handle the cropping instead of passing to downstream.

wenqingx:
thanks for review, and Focus help to do this part.closed. https://github.com/intel/vaapi-fits/pull/522

PR #526, opened 2023-06-21
Correct HDR parameters

xhaihao:


xhaihao:
@uartie @Bin-CI  May I merge this PR ?

xhaihao:
> LGTM, but the way compare two list is too complex how about `sorted(list_a) == sorted(list_b)`

Here compare the first element of the list only, we needn't compare the list.

PR #532, opened 2023-07-03
lib/ffmpeg/transcoderbase: show all frame's HDR info

feiwan1:
In case of input has audio stream which will influence frame number of output video. The other reason is that vframes for filter showinfo is not accurate. For example, when use 'vframes 1', the output of showinfo will be 2 frames' info. Use all frame number to make sure the otuput will be same for diffferent case.

### DIFF
--- a/lib/ffmpeg/transcoderbase.py
+++ b/lib/ffmpeg/transcoderbase.py
@@ -271,7 +271,7 @@ class BaseTranscoderTest(slash.Test,BaseFormatMapper):
 
   def get_hdr_info(self, osfile):
     output = call(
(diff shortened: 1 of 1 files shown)


xhaihao:
Is the performance impacted when the frame number is large ?

feiwan1:
> Is the performance impacted when the frame number is large ?



> Is the performance impacted when the frame number is large ?

Re-pushed to disable audio stream. Which fix the input has different video frames comapre with output.

xhaihao:
@feiwan1 Please fix errors in pre-check.

feiwan1:
Hi @Bin-CI, could you help to check container build fail issue in pr check? Thanks.

PR #543, opened 2023-08-28
Add expand context disp_win

mengker33:
Add a new expand conetxt disp_win to handle the hevc decoding cases which contain disp_win, this is because vpl and vaapi uses disp_win offsets differently and they are expexted.


.slashrc
```diff
@@ -224,6 +226,9 @@ class MediaPlugin(slash.plugins.PluginInterface):
     from lib.platform import info
     return info()["os"]
 
+  def _get_component(self):
+    return sys.argv[2].split("/")[1]
+
```
uartie, line 231: sys.argv does not always contain the component.  Users can run all tests by only specifying  `run test/` directory, for example

uartie:
unnecessary

PR #550, opened 2023-09-12
Support trendline models for AV1e 10bit

uartie:


FocusLuo:
@uartie there is a conflict need to be fixed

PR #551, opened 2023-09-14
[Trend Mode] add Vp9 8bit trendline support

FocusLuo:
1, add ffmpeg-qsv vp9 8bit profile map
2, move ffmpeg-qsv vp9-8 encode test classes to lib
3, add support for vp9 8bit encode trendline models

### DIFF
--- a/model/encode/vp9.py
+++ b/model/encode/vp9.py
@@ -1,0 +1,44 @@
+###
+### Copyright (C) 2023 Intel Corporation
+###
+### SPDX-License-Identifier: BSD-3-Clause
+###
+
+import os
+import slash
+
+from ...lib.ffmpeg.qsv.util import *
+from ...lib.ffmpeg.qsv.encoder import VP9_8EncoderLPTest
+from .util import TrendModelMixin
+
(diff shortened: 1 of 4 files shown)


model/encode/vp9.py
```diff
@@ -0,0 +1,39 @@
...
+    if "AYUV" == self.format:
+      profile = "profile1"
+
+    super().initvars(profile)
+
+  @slash.parametrize(*TrendModelMixin.filter_spec(spec))
+  def test(self, case):
+    vars(self).update(case = case)
+    vars(self).update(spec[case].copy())
+    self.fit()
```
uartie, line 35: We should use the same "modelqps" settings as we do in AV1 because VP9 has QP range 1..255.  Unfortunately, due to open bug, we'll have to modify ffmpeg-qsv to allow QP > 51 in order to generate the correct trend models.

uartie (changes requested):

model/encode/vp9.py
FocusLuo: thanks, resolved

uartie (approved):

PR #557, opened 2023-09-21
[MTL] non-low power mode support for vp9 8bit encode

FocusLuo:
there will no lp mode for MTL encode
add non-lp mode support for vp9 8bit encode for ffmpeg and gstreamer

PR #561, opened 2023-09-25
[MTL+] add av1-8bit non low power mode support

FocusLuo:
add av1-8bit non low power mode support for MTL+ platforms

### DIFF
--- a/test/ffmpeg-qsv/encode/av1.py
+++ b/test/ffmpeg-qsv/encode/av1.py
@@ -6,10 +6,96 @@
 
 from ....lib import *
 from ....lib.ffmpeg.qsv.util import *
-from ....lib.ffmpeg.qsv.encoder import AV1EncoderLPTest
+from ....lib.ffmpeg.qsv.encoder import AV1EncoderLPTest, AV1EncoderTest
 
 spec = load_test_spec("av1", "encode", "8bit")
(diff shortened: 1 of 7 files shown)


uartie (changes requested):
I have same concerns as https://github.com/intel/vaapi-fits/pull/557

test/ffmpeg-qsv/encode/av1.py
```diff
@@ -6,11 +6,97 @@
...
+      bframes   = bframes,
+      qp        = qp,
+      rcmode    = "cqp",
+      quality   = quality,
+      profile   = profile,
+      tilerows  = tilerows,
+      tilecols  = tilecols,
+    )
+
+  @slash.parametrize(*gen_av1_cqp_lp_parameters(spec))
```
uartie, line 29: This is the wrong parameter generator... it is getting config from the cqp_lp variant.  Please add a new generator that generates cases from the "cqp" variants.

Also, please do the same for the others (cbr, vbr).

uartie (changes requested):
wrong case variant generators.

test/ffmpeg-qsv/encode/av1.py
FocusLuo: updated

uartie (approved):
LGTM

PR #565, opened 2023-09-27
Mtl parametrize vp9 8bit enc

FocusLuo:
MTL replace non-low power encode parameters for ffmpeg/gstreamer
gen_vp9_cqp_lp_parameters -> gen_vp9_cqp_parameters
gen_vp9_cbr_lp_parameters -> gen_vp9_cbr_parameters
gen_vp9_vbr_lp_parameters -> gen_vp9_vbr_parameters

### DIFF
--- a/lib/parameters.py
+++ b/lib/parameters.py
@@ -612,17 +612,17 @@ def gen_vp9_cqp_variants(spec):
     variants = params.get("variants", dict()).get("cqp", None)
 
     if variants is None:
-      keys = ["ipmode", "qp", "quality", "refmode", "looplvl", "loopshp"]
-      product = itertools.product([0, 1], [14, 28], [1, 4, 7], [0, 1], [0, 16, 63], [0, 4, 7])
+      keys = ["ipmode", "qp", "quality", "slices", "refmode", "looplvl", "loopshp"]
+      product = itertools.product([0, 1], [14, 28], [1, 4, 7], [1, 3], [0, 1], [0, 16, 63], [0, 4, 7])
       variants = [dict(zip(keys, vals)) for vals in product]
 
     for variant in variants:
       yield [
-        case, variant["ipmode"], variant["qp"], variant["quality"],
+        case, variant["ipmode"], variant["qp"], variant["quality"], variant.get("slices", 1),
         variant["refmode"], variant["looplvl"], variant["loopshp"]]
 
 def gen_vp9_cqp_parameters(spec):
-  keys = ("case", "ipmode", "qp", "quality", "refmode", "looplvl", "loopshp")
+  keys = ("case", "ipmode", "qp", "quality", "slices", "refmode", "looplvl", "loopshp")
   params = gen_vp9_cqp_variants(spec)
   return keys, params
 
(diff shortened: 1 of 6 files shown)


FocusLuo:
BTW, because iHD supported slices for ffmpeg-qsv/gst-msdk, so in lib add "slices" for vp9 cqp/cbr/vbr encode. but this should impact i965 ffmpeg-vaapi/gst-vaapi test cases reference name on Gen9.5- platforms, so we need also update i965 ffmpeg-vaapi/gst-vaapi test reference

test/gst-vaapi/encode/vp9.py
```diff
@@ -82,13 +82,13 @@ def init(self, tspec, case, gop, bitrate, fps, refmode, looplvl, loopshp):
       refmode   = refmode,
     )
  
-  @slash.parametrize(*gen_vp9_cbr_parameters(spec))
-  def test(self, case, gop, bitrate, fps, refmode, looplvl, loopshp):
-    self.init(spec, case, gop, bitrate, fps, refmode, looplvl, loopshp)
+  @parametrize_with_unused(*gen_vp9_cbr_parameters(spec), ['slices'])
+  def test(self, case, gop, bitrate, fps, slices, refmode, looplvl, loopshp):
+    self.init(spec, case, gop, bitrate, fps, slices, refmode, looplvl, loopshp)
```
uartie, line 87: If slices is unused, then it does not need to be passed to "init"

uartie:
> BTW, because iHD supported slices for ffmpeg-qsv/gst-msdk, so in lib add "slices" for vp9 cqp/cbr/vbr encode. but this should impact i965 ffmpeg-vaapi/gst-vaapi test cases reference name on Gen9.5- platforms, so we need also update i965 ffmpeg-vaapi/gst-vaapi test reference

Since we're using trendlines for encode, there should not be any test reference names based on the parameter list anymore.

uartie (changes requested):

lib/parameters.py
```diff
@@ -612,17 +612,17 @@ def gen_vp9_cqp_variants(spec):
...
-      keys = ["ipmode", "qp", "quality", "refmode", "looplvl", "loopshp"]
-      product = itertools.product([0, 1], [14, 28], [1, 4, 7], [0, 1], [0, 16, 63], [0, 4, 7])
+      keys = ["ipmode", "qp", "quality", "slices", "refmode", "looplvl", "loopshp"]
+      product = itertools.product([0, 1], [14, 28], [1, 4, 7], [1, 3], [0, 1], [0, 16, 63], [0, 4, 7])
       variants = [dict(zip(keys, vals)) for vals in product]
 
     for variant in variants:
       yield [
-        case, variant["ipmode"], variant["qp"], variant["quality"],
+        case, variant["ipmode"], variant["qp"], variant["quality"], variant["slices"],
```
uartie, line 621: Please make slices optional with a default for user config.  We already have some cases defined without slices.

`variant.get("slices", 1)`

Do same for the other vp9 generators, please.

uartie: This is why the pre-check failed

FocusLuo: updated

test/gst-vaapi/encode/vp9.py
FocusLuo: updated

uartie (approved):
LGTM

PR #569, opened 2023-10-04
Allow Slash 1.14

uartie:
Newer version of setuptools/pip throw error when trying
to install slash 1.12.0 due to more strict dependency
version syntax checks.

The syntax was fixed in slash 1.14.0 (slash commit 6ceffde64970).

Tested vaapi-fits with slash 1.14.0 and it still works well.


FocusLuo (approved):
LGTM

PR #571, opened 2023-10-18
[requirement] add matplotlib

FocusLuo:
add matplotlib which be used by trend mode


uartie (approved):
LGTM.  Btw, it's not strictly required for normal testing and CI.  Only needed if user wants to generate trendline models.  Since CI does not pull requirements from this file, this change is ok.

PR #574, opened 2023-10-20
Replace hard-coded strings in "codec" member properties

uartie:
Depends on https://github.com/intel/vaapi-fits/pull/573

Different and inconsistent hard-coded strings that represent
the same codec can introduce bugs and make it harder to maintain.
Thus, use the new common Codec enum to improve consistency and
maintenance problems.
    
The bitdepth in codec strings are overkill and unnecessary, so
that distinction is removed.
    
For transcode, currently only 8-bit is supported and assumed.
Additional bitdepth support may be added in the future using
a different logic.

### DIFF
--- a/lib/ffmpeg/encoderbase.py
+++ b/lib/ffmpeg/encoderbase.py
@@ -15,7 +15,7 @@ from ...lib.ffmpeg.decoderbase import Decoder
 from ...lib.parameters import format_value
 from ...lib.properties import PropertyHandler
 from ...lib.util import skip_test_if_missing_features
-
+from ...lib.codecs import Codec
 from ...lib import metrics2
 
 class Encoder(PropertyHandler, BaseFormatMapper):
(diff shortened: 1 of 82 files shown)


uartie:
> LGTM, one question: if we merge the `hevc-*` into `hevc` codec; `vp9-*` into `vp9`. How about the case of this part, we also need to merge them?

The test spec keys (e.g. `media._get_test_spec("hevc", "encode", "10bit")` and `media._get_test_spec("hevc", "encode", "8bit")`) are still treated separately.  So there is no change there.
However, we may be able to consolidate/merge the test base classes now (e.g. `HEVC8EncoderBaseTest`, `HEVC10EncoderBaseTest` and `HEVC12EncoderBaseTest` are the same now).  I will take care of this in another PR.

uartie:
> > LGTM, one question: if we merge the `hevc-*` into `hevc` codec; `vp9-*` into `vp9`. How about the case of this part, we also need to merge them?
> 
> The test spec keys (e.g. `media._get_test_spec("hevc", "encode", "10bit")` and `media._get_test_spec("hevc", "encode", "8bit")`) are still treated separately. So there is no change there. However, we may be able to consolidate/merge the test base classes now (e.g. `HEVC8EncoderBaseTest`, `HEVC10EncoderBaseTest` and `HEVC12EncoderBaseTest` are the same now). I will take care of this in another PR.

@Bin-CI consolidation/merge is here: https://github.com/intel/vaapi-fits/pull/577

PR #577, opened 2023-11-01
Move and consolidate codec-specialized encode test classes to lib

uartie:
Move all the codec-specialized encode test classes to lib so that they can be reused.  Use dynamic class template to generate test classes.

### DIFF
--- a/.slashrc
+++ b/.slashrc
@@ -56,6 +56,7 @@ class MediaPlugin(slash.plugins.PluginInterface):
   testspec = dict()
 
   platform = None
+  render_device = "/dev/dri/renderD128"
   suite = os.path.basename(sys.argv[0])
   mypath = __SCRIPT_DIR__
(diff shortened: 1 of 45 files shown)


uartie:
> How about transcode?

Transcode tests are currently out of scope for this change.  I am trying to improve/cleanup encode tests first so I can add support for encoded/compressed **input** file assets.

uartie:
> > > How about transcode?
> > 
> > 
> > Transcode tests are currently out of scope for this change. I am trying to improve/cleanup encode tests first so I can add support for encoded/compressed **input** file assets.
> 
> I'm ok for this, just curious because `transcode` will load `decode` and `encode` part, so if `encode` have the big change, whether it should been worked between different workload

Yeh, transcode tests do not share any code with the encode/decode tests yet.  Hopefully in the future.

PR #581, opened 2023-11-14
[Icq] gst-msdk added av1-8bit encode ICQ mode support

FocusLuo:
1, LP AV1 ICQ mode
2, Non-LP ICQ mode
3, gst-msdk cmd line change

### DIFF
--- a/lib/gstreamer/msdk/encoder.py
+++ b/lib/gstreamer/msdk/encoder.py
@@ -41,6 +41,9 @@ class Encoder(GstEncoder):
       if self.codec in [Codec.MPEG2]:
         mqp = mapRangeInt(qp, [0, 100], [0, 51])
         return f" qpi={mqp} qpp={mqp} qpb={mqp}"
+      if self.codec in [Codec.AV1]:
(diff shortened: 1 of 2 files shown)


uartie (approved):
LGTM.

How do we enable `msdkav1enc` element in gst-msdk?  I don't see it on my DG2.  Also, it does not appear the CI detects it either.

FocusLuo:
> msdkav1enc
On DG2/ATSM, iHD uses LP av1 enc, so gst-msdk only has msdkav1lpenc 
On MTL, iHD uses Non-LP av1 enc, so gst-msdk only has msdkav1enc

uartie:
> > msdkav1enc
> > On DG2/ATSM, iHD uses LP av1 enc, so gst-msdk only has msdkav1lpenc
> > On MTL, iHD uses Non-LP av1 enc, so gst-msdk only has msdkav1enc

gst-msdk does not have separate elements for lp vs. non-lp.  I think you got it confused with gst-va.
Anyhow, I figured out the problem (see VIZ-20142)

PR #598, opened 2023-12-04
[ffmpeg-qsv] Add VPP CSC 10bit test suites

Hangjie22Coder:


uartie (changes requested):
I don't think we need this 10bit specialized test.  We can use the existing vpp/csc.py to test 10bit formats.  Just add it to the test case config with explicit 10bit formats.

uartie:
> I don't think we need this 10bit specialized test. We can use the existing vpp/csc.py to test 10bit formats. Just add it to the test case config with explicit 10bit formats.

The only thing that this specialization will do is determine how it is reported by our current report generators.  But I don't think that is enough to warrant the addition.

Hangjie22Coder:
Thanks for the feedback! I agree with you that don't add specialized tests for the 10-bit format. I'll update the PR to use the existing vpp/csc .py tests.

PR #599, opened 2023-12-05
gst/decode: auto select demux element

uartie:
Try to auto select the gstreamer demuxer element from the filename extension for some of the most common container streams.


FocusLuo (approved):
LGTM

PR #600, opened 2023-12-06
Support encoded source/input files for Encode/VPP

uartie:
User config may specify an encoded "source" file (e.g. as AVC, HEVC, ...) to be used as input to encode or vpp test cases.  The source file encoding should be specified in the test case definition with "scodec" property (e.g. Codec.AVC, Codec.HEVC, ...).  If the encoding is not specified then Codec.RAW (e.g. yuv/rgb) is assumed.

When using an encoded source file as input, it will be decoded internally to a temporary RAW file (using the "format" specified in the case definition).  This temporary RAW file will then be used in the usual encode/vpp command pipeline.  The temporary RAW files are cached for the entire test session and therefore the source file only needs decoded once and can be used multiple times by other test cases without re-decoding.  The temporary RAW files are deleted after the test run/session finishes (depending on retention option).

### DIFF
--- a/lib/ffmpeg/qsv/vpp.py
+++ b/lib/ffmpeg/qsv/vpp.py
@@ -7,6 +7,7 @@
 import slash
 
 from ....lib.common import get_media, mapRangeInt, mapRangeWithDefault
+from ....lib.ffmpeg.qsv.decoder import Decoder
 from ....lib.ffmpeg.qsv.util import using_compatible_driver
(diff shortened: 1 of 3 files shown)


uartie:
> LGTM, just one question, if case failed, how to detail the `source` it will been cleanup at which time?

The decoded raw source is stored and cleaned up at "SESSION" scope.  That is, it  will be stored in the top-level results directory for the entire test run session and removed according to retention option after the session ends.  You can set the `--artifact-retention` option appropriately (see command help) to prevent the file from being deleted after the run.

PR #606, opened 2023-12-13
[hevc qvbr] ffmpeg-vaapi and qsv add hevc 8bit qvbr encode support

FocusLuo:
[hevc qvbr] ffmpeg-vaapi and qsv add hevc 8bit qvbr encode support


uartie (approved):
LGTM

PR #607, opened 2023-12-14
ffmpeg/xcode: replace vsync with fps_mode

uartie:
ffmpeg vsync option renamed to fps_mode


uartie:
Possible test regression.  Investigating...

uartie:
> Possible test regression. Investigating...

The `vsync` option is applied globally, but is deprecated.  The `fps_mode` only applies to one stream.  Thus, need to add an `fps_mode` per xcode output.  Updating...

PR #609, opened 2023-12-18
test/ffmpeg/gstreamer: add vpp/csc/10bit r2r test

Hangjie22Coder:


uartie (changes requested):
Similar comments as #598... we don't need 10bit specialization.

PR #612, opened 2023-12-21
[qvbr] add gst-msdk hevc-8 qvbr encode test implement

FocusLuo:
add gst-msdk hevc-8 qvbr encode test implement


FocusLuo:
gst-va doesn't support hevc qvbr yet, so this PR only includes for gst-msdk

uartie (approved):
LGTM

PR #614, opened 2023-12-22
ffmpeg/decode: fix autoscale usage break

uartie:
The -autoscale 0 option no longer works and causes pipeline to error/abort.

The correct usage, apparently, is to use -noautoscale.
```
Since FFMpeg commit 25c98566e8a45b50415c80ca6450282f2ec0657a
Author: Anton Khirnov <anton@khirnov.net>
Date:   Fri Dec 15 02:31:03 2023 +0100

    fftools/ffmpeg_opt: drop HAS_ARG from auto{scale,rotate}
```


uartie:
cc: @wenqingx

uartie:
Emergency merge after pre-check since all ffmpeg dec/enc tests are broken without this.

PR #619, opened 2024-01-05
[vpp/stack] variable metric check 

wenqingx:


wenqingx:
@feiwan1  suggest should be use ssim to check stack. 

please help a reivew ,if no problem will update the new baseline yuv file to server.

uartie:
SSIM is slow and will require us to store YUV reference files.

IIRC, it's only an issue between windows and linux differences.  How about we use different MD5 ref for Windows, instead?

uartie (changes requested):
no reference handling?

wenqingx:
> handling?

I haven't submitted the refs YUV yet. If you think this patch is acceptable, I will find Bin to help update it.

feiwan1:
Md5 is different on different platform for linux. It's OK for me to use md5 method. If use md5, we need to use different md5 for each platforms.

wenqingx:
> SSIM is slow and will require us to store YUV reference files.
> 
> IIRC, it's only an issue between windows and linux differences. How about we use different MD5 ref for Windows, instead?

I understand. My patch is not the same issue as what you previously learned.
You are referring to the different issues with MD5 on different operating systems (Linux and Windows) before.

I am here to fix the issue of different MD5 values on different platforms under the same Linux operating system. Is it acceptable to use different MD5 reference values on different platforms under the same OS?

uartie:
> I am here to fix the issue of different MD5 values on different platforms under the same Linux operating system. Is it acceptable to use different MD5 reference values on different platforms under the same OS?

Hmmm... that would be too much overhead to maintain.  I suppose ssim is fine in this case as long as it is well documented in the test case config.

lib/mixin/vpp.py
```diff
@@ -164,10 +164,10 @@ def check_stack(self):
       params.update(width = self.cols * self.tilew, height = self.rows * self.tileh)
       params.update(reference = format_value(self.reference, **params))
     elif self.stack in ["hstack"]:
-      params.update(metric = dict(type = "md5"))
+      params.update(metric = dict(type = "ssim", miny = 0.97, minu = 0.97, minv = 0.97))
```
uartie, line 167:
```suggestion
      params.setdefault(metric = "md5")
```

lib/mixin/vpp.py
```diff
@@ -164,10 +164,10 @@ def check_stack(self):
...
       params.update(width = self.inputs * self.width, height = self.height)
     elif self.stack in ["vstack"]:
-      params.update(metric = dict(type = "md5"))
+      params.update(metric = dict(type = "ssim", miny = 0.97, minu = 0.97, minv = 0.97))
```
uartie, line 170:
```suggestion
      params.setdefault(metric = "md5")
```

lib/mixin/vpp.py
```diff
@@ -164,10 +164,10 @@ def check_stack(self):
...
       params.update(width = self.width, height = self.height * self.inputs)
     else:
       assert False, f"unknown stack operation: '{self.stack}'"
```
uartie, line 173:
```suggestion
    else:
      assert False, f"unknown stack operation: '{self.stack}'"
    
    # a reference is needed by some metrics
    if vars(self).get("reference", None) is not None:
      params.update(reference = format_value(self.reference, **params))
```

lib/mixin/vpp.py
```diff
@@ -164,10 +164,10 @@ def check_stack(self):
       params.update(width = self.cols * self.tilew, height = self.rows * self.tileh)
       params.update(reference = format_value(self.reference, **params))
```
uartie, line 165:
```suggestion
      
```

uartie (changes requested):
If we change the default metric then we'll break _all_ of the test cases for _h/vstack_ everywhere.

Instead, let's keep the default and make it user configurable.

You commit message says "xstack" but this patch does not change xstack.  I think you mean "vstack"

wenqingx:
> > I am here to fix the issue of different MD5 values on different platforms under the same Linux operating system. Is it acceptable to use different MD5 reference values on different platforms under the same OS?
> 
> Hmmm... that would be too much overhead to maintain. I suppose ssim is fine in this case as long as it is well documented in the test case config.

Okay, if you agree, I will try using SSIM for inspection and test the results on multiple platforms.

lib/mixin/vpp.py
wenqingx: Okay, I understand what you mean. I'm trying to make changes now,thanks.

lib/mixin/vpp.py
uartie: no, it should not be necessary to modify this.

wenqingx: ok,thanks, will update the user case.

lib/mixin/vpp.py
```diff
@@ -162,16 +162,19 @@ def check_stack(self):
     if self.stack in ["xstack"]:
       params.update(metric = dict(type = "ssim", miny = 0.97, minu = 0.97, minv = 0.97))
```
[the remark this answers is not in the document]
uartie: Yes, we can change this one too.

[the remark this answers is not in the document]
wenqingx: alsp updated

lib/mixin/vpp.py
```diff
@@ -160,18 +160,21 @@ def check_stack(self):
...
-      params.update(metric = dict(type = "md5"))
+      params.setdefault("metric", dict(type = "md5"))
       params.update(width = self.width, height = self.height * self.inputs)
     else:
       assert False, f"unknown stack operation: '{self.stack}'"
 
+    # a reference is needed by some metrics
+    if vars(self).get("reference", None) is not None:
+      params.update(reference = format_value(self.reference, **params))
+
```
uartie, line 177: Hmm, I guess this won't work with our test configs b/c the reference template assumes xstack :/  Hence the check failures.  We need to think of another way to handle this.  I will try to take look into it next week.

uartie (changes requested):

lib/mixin/vpp.py
uartie: Maybe we can use the post validate in test config to set the reference based on stack mode.

uartie: Ok, I see you've updated the test config

uartie: Please send patches for bat and full test configs.

wenqingx: ok , will update these part, thanks for your help and review.

uartie (approved):

PR #625, opened 2024-01-26
HEVC 8bit 422 encode support

FocusLuo:
Added HEVC 8bit 422 encode support for ffmpeg-vaapi/qsv and gst-va/vaapi/msdk

### DIFF
--- a/lib/caps/ADL/iHD
+++ b/lib/caps/ADL/iHD
@@ -33,7 +33,7 @@ caps = dict(
     mpeg2   = dict(maxres = res2k , fmts = ["NV12"]),
     hevc_8  = dict(
       maxres    = res8k,
-      fmts      = ["NV12", "AYUV"],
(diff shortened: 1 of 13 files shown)


uartie:
I admit I was a little confused at first that 8bit 422 was being mapped to 10bit 422 profiles.  But I guess that is correct according to https://en.wikipedia.org/wiki/High_Efficiency_Video_Coding#Profiles

Can you add some comments in the code about this so others don't encounter the same confusion?

FocusLuo:
> I admit I was a little confused at first that 8bit 422 was being mapped to 10bit 422 profiles. But I guess that is correct according to https://en.wikipedia.org/wiki/High_Efficiency_Video_Coding#Profiles
> 
> Can you add some comments in the code about this so others don't encounter the same confusion?

Right, 8bit hevc 422 profile is mapped to main-422-10, based on hevc spec define

uartie (approved):

PR #629, opened 2024-02-20
[ffmpeg-qsv] add encode/hevc playback_seek test case

Hangjie22Coder:


test/ffmpeg-qsv/encode/hevc.py
```diff
@@ -175,7 +175,7 @@ def test_r2r(self, case, gop, slices, bitrate, fps, quality, refs, profile):
 
   # TCBRC is VBR LP + LDB + Strict(-1)
   @slash.parametrize(*gen_hevc_tcbrc_parameters(spec, ['main']))
-  def test_tcbrc(self, case, bitrate, fps, profile):
```
uartie, line 178: Changing an existing test function signature/params is risky and generally not acceptable.  It can break other component tests if not carefully done correctly.  It also messes up CI test run histories.

It would be better to create a new test function and parameter generator to handle seek cases specifically, e.g. `test_tcbrc_seek`.

uartie (changes requested):

test/ffmpeg-qsv/encode/hevc.py
uartie: Also, is it necessary to test seek with all the different encode combinations (e.g. vbr, cbr, tcbrc, etc..)?  Maybe we only need a simple seek test that uses middleware defaults?

lib/ffmpeg/qsv/encoder.py
```diff
@@ -92,6 +93,7 @@ def check_metrics(self):
       vars(self).get("lowpower", 0),
       vars(self).get("ldb", 0),
       vars(self).get("strict", 0) == -1,
+      vars(self).get("seek", 0)
```
uartie, line 96: Hmm... why does `tcbrc` strictly require a seek option?  I mean, we should be able to do `tcbrc` without `seek`, right?

Hangjie22Coder: Yes, but because we lack seek-related test cases, I added this case (Recode_peru_8K_HDR), I think all modes of this case (e.g. CBR/VBR/CQP) need to be tested for seek, and currently the mode of tcbrc is submitted first.

Hangjie22Coder:
Hi @uartie Can you review this patch for me?

lib/parameters.py
```diff
@@ -58,6 +58,42 @@ def gen_avc_cqp_parameters(spec, profiles):
...
 
+def gen_avc_cqp_seek_variants(spec, profiles):
+  for case, params in spec.items():
+    variants = copy.deepcopy(params.get("variants", dict()).get("cqp", None))
+    if variants is None:
+      keys = ["gop", "slices", "bframes", "qp", "quality", "profile"]
+      product  = list(itertools.product([1], [1], [0], [14, 28], [1, 4, 7], profiles))  # I, single-slice
+      product += list(itertools.product([30], [1], [0], [14, 28], [1, 4, 7], profiles)) # IP, single-slice
+      product += list(itertools.product([30], [4], [2], [14, 28], [1, 4, 7], profiles)) # IPB, multi-slice
+      variants = [dict(zip(keys, vals)) for vals in product]
```
uartie, line 69: We don't need defaults generated if the variant is not defined in test config.

Also, we should not reuse the "cqp" variant key.  It is meant for the "cqp" tests and has a different parameter signature.  If we reuse it, then there will suddenly be double amount of cases generated from current test configs (a set for cqp.test and a set for cqp_seek.test) that do the exact same thing.  

Same concern for "vbr" and "tcbr" added here too.

test/ffmpeg-qsv/encode/hevc.py
```diff
@@ -60,6 +60,26 @@ def test_r2r(self, case, gop, slices, qp, quality, profile):
...
+      rcmode  = "cqp",
+      slices  = slices,
+      seek    = seek,
+    )
+
+  @slash.parametrize(*gen_hevc_cqp_seek_parameters(spec, ['main']))
+  def test(self, case, gop, slices, bframes, qp, quality, profile, seek):
+    self.init(spec, case, gop, slices, bframes, qp, quality, profile, seek)
+    self.encode()
+
```
uartie, line 82: Maybe it would be better to use similar test signature like the `pict` test classes.  This way, we would not need different classes for cqp, cbr, tcbrc, and vbr variations.  That is, the rcmode can be passed as a parameter and perhaps integrate tcbrc params into one test definition.

uartie (changes requested):
Sorry, I missed the review request... I don't always see the "draft" PRs right away.

test/ffmpeg-qsv/encode/hevc.py
```diff
@@ -282,7 +283,7 @@ def test(self, case, bitrate, maxrate, fps, maxframesize, profile):
     self.encode()
 
 class pict(HEVC8EncoderTest):
-  def init(self, tspec, case, gop, bframes, bitrate, qp, maxrate, profile, rcmode):
```
uartie, line 285: I didn't mean to imply we should change the `pict` test class.  Rather, we should copy it and name it for a new `seek` test class.

uartie (changes requested):

test/ffmpeg-qsv/encode/hevc.py
```diff
@@ -363,3 +363,46 @@ def test(self, case, gop, bframes, bitrate, maxrate, profile, rcmode):
...
+      rcmode     = rcmode,
+      bframes    = bframes,
+      bitrate    = bitrate,
+      case       = case,
+      maxrate    = maxrate,
+      minrate    = bitrate,
+      profile    = profile,
+      qp         = qp,
+      gop        = gop,
+      vseek      = 1,
```
uartie, line 379: what is "vseek" used for?

uartie (changes requested):

test/ffmpeg-qsv/encode/hevc.py
Hangjie22Coder: Sorry I ignored this, it's been updated.

lib/ffmpeg/qsv/encoder.py
```diff
@@ -63,6 +63,7 @@ def encparams(self):
       f"{super().encparams}{self.ldb}"
       f"{self.iqfactor}{self.bqfactor}"
       f"{self.iqoffset}{self.bqoffset}"
+      f"{self.seek}"
```
uartie, line 66: Since this property is in the base class and is generic, please add this part to the base class `encparams`, instead.

uartie (changes requested):

uartie:
Unfortunately, this code is not enough to handle seek cases.  That is, many checks depend on "frames" user config parameter and will fail.  For example, the bitrate gap check will fail because we don't compensate for the "skipped" frames caused by the seek option.  That is, if we encode a source file that has 150 frames @25FPS and set seek to 5, then the encoded output file will only be 25 frames.

For `seek`, ffmpeg documents it as `-ss <time_off> start transcoding at specified time`. 

Therefore, we need to figure out a good approach to handle this appropriately.

uartie:
> Unfortunately, this code is not enough to handle seek cases. That is, many checks depend on "frames" user config parameter and will fail. For example, the bitrate gap check will fail because we don't compensate for the "skipped" frames caused by the seek option. That is, if we encode a source file that has 150 frames @ 25FPS and set seek to 5, then the encoded output file will only be 25 frames.
> 
> For `seek`, ffmpeg documents it as `-ss <time_off> start transcoding at specified time`.
> 
> Therefore, we need to figure out a good approach to handle this appropriately.

For other checks, like PSNR or SSIM, the reference will need to "seek" before frames are compared with the output, also.  Now, the complexity of this task is much harder.

uartie:
It would be good to provide a test case, too, to verify it works as intended.

uartie:
@Hangjie22Coder I submitted a PR to your branch that links to this PR... you can merge it to your branch and it should automatically update here.

Hangjie22Coder:
> I submitted a PR to your branch that links to this PR... you can merge it to your branch and it should automatically update here.

Thanks for submitting a PR to link to mine!I've merged it into my branch. Thanks for your help!

test/ffmpeg-qsv/encode/hevc.py
```diff
@@ -363,3 +363,44 @@ def test(self, case, gop, bframes, bitrate, maxrate, profile, rcmode):
...
+      maxrate    = maxrate,
+      minrate    = bitrate,
+      profile    = profile,
+      qp         = qp,
+      gop        = gop,
+      seek       = seek,
+    )
+
+  @slash.parametrize(*gen_hevc_seek_parameters(spec, ['main']))
+  def test(self, case, gop, bframes, bitrate, qp, maxrate, profile, rcmode, seek):
```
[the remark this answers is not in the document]
uartie: @Hangjie22Coder can you add "fps" as a parameter, too?  Or do you need help with that?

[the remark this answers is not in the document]
Hangjie22Coder: Sorry I ignored this, it's been updated.

uartie:
It looks like you added my commit as a "merge commit".  It needs to be added with a "rebase and merge" commit please.

lib/parameters.py
```diff
@@ -516,6 +516,70 @@ def gen_hevc_pict_lp_parameters(spec, profiles):
...
+def gen_hevc_seek_variants(spec, profiles):
+  for case, params in spec.items():
+    for variant in copy.deepcopy(params.get("variants", dict()).get("seek", [])):
+      uprofile = variant.get("profile", None)
+      cprofiles = [uprofile] if uprofile else profiles
+      gop     = variant.get("gop", None)
+      bframes = variant.get("bframes", None)
+      bitrate = variant.get("bitrate", None)
+      qp      = variant.get("qp", None)
+      fps     = variant.get("fps", None)
```
uartie, line 528:
```suggestion
      fps     = variant.get("fps", 25)
```

lib/parameters.py
```diff
@@ -516,6 +516,70 @@ def gen_hevc_pict_lp_parameters(spec, profiles):
...
+def gen_hevc_seek_lp_variants(spec, profiles):
+  for case, params in spec.items():
+    for variant in copy.deepcopy(params.get("variants", dict()).get("seek_lp", [])):
+      uprofile = variant.get("profile", None)
+      cprofiles = [uprofile] if uprofile else profiles
+      gop     = variant.get("gop", None)
+      bframes = variant.get("bframes", None)
+      bitrate = variant.get("bitrate", None)
+      qp      = variant.get("qp", None)
+      fps     = variant.get("fps", None)
```
uartie, line 560:
```suggestion
      fps     = variant.get("fps", 25)
```

uartie (changes requested):

uartie:
Please squash your commits and remove the "merge" commit.  There should only be 2 commits, total (your commit and my commit).

uartie (approved):

PR #635, opened 2024-04-03
Add support for FFmpeg-Vaapi pad

xhaihao:


lib/ffmpeg/vaapi/vpp.py
```diff
@@ -42,6 +42,14 @@ def gen_vpp_opts(self):
...
       vpfilter.append("[1:v]hwupload[v1];")
       vpfilter[-1] += "[0:v][v1]overlay_vaapi=alpha={a}".format(a = mapRange(self.alpha, [0.0, 255.0], [0.0, 1.0]))
+    elif self.vpp_op in ["pad"]:
+      vpfilter.append("pad_vaapi")
+      if hasattr(self, "padw") and hasattr(self, "padh"):
+        vpfilter[-1] += f"=width={self.padw}:height={self.padh}"
+        if hasattr(self, "x") and hasattr(self, "y"):
+          vpfilter[-1] += f":x={self.x}:y={self.y}"
+          if hasattr(self, "color"):
+            vpfilter[-1] += f":color={self.color}"
```
uartie, line 52: What about `color` setting without `x` and `y`?

lib/ffmpeg/vaapi/vpp.py
```diff
@@ -42,6 +42,14 @@ def gen_vpp_opts(self):
     elif self.vpp_op in ["overlay"]:
       vpfilter.append("[1:v]hwupload[v1];")
       vpfilter[-1] += "[0:v][v1]overlay_vaapi=alpha={a}".format(a = mapRange(self.alpha, [0.0, 255.0], [0.0, 1.0]))
+    elif self.vpp_op in ["pad"]:
+      vpfilter.append("pad_vaapi")
+      if hasattr(self, "padw") and hasattr(self, "padh"):
+        vpfilter[-1] += f"=width={self.padw}:height={self.padh}"
+        if hasattr(self, "x") and hasattr(self, "y"):
```
uartie, line 49: We have to have both `x` and `y`?  Why not allow just `x` or just `y`?

lib/ffmpeg/vaapi/vpp.py
```diff
@@ -42,6 +42,14 @@ def gen_vpp_opts(self):
     elif self.vpp_op in ["overlay"]:
       vpfilter.append("[1:v]hwupload[v1];")
       vpfilter[-1] += "[0:v][v1]overlay_vaapi=alpha={a}".format(a = mapRange(self.alpha, [0.0, 255.0], [0.0, 1.0]))
+    elif self.vpp_op in ["pad"]:
+      vpfilter.append("pad_vaapi")
+      if hasattr(self, "padw") and hasattr(self, "padh"):
+        vpfilter[-1] += f"=width={self.padw}:height={self.padh}"
```
uartie, line 48: We have to have both `padw` and `padh`?  Why not allow just `padw` or just `padh`?

lib/ffmpeg/vaapi/vpp.py
uartie, line 52:
```suggestion
      vpopts = []
      if hasattr(self, "padw"): vpopts.append(f"width={self.padw}")
      if hasattr(self, "padh"): vpopts.append(f"height={self.padh}")
      
      # ...and so on...
      
      if len(vpopts): vpfilter[-1] += f"={':'.join(vpopts)}"
```

uartie (changes requested):

lib/ffmpeg/vaapi/vpp.py
xhaihao: There are no dependency between `x` and `y`,  I'll follow your suggestion to update it.

lib/ffmpeg/vaapi/vpp.py
xhaihao: There are no dependency between `padw` and `padh`, I'll follow your suggestion to update it.

lib/ffmpeg/vaapi/vpp.py
xhaihao: Updated, thx

lib/ffmpeg/vaapi/vpp.py
xhaihao: Updated, thx

lib/ffmpeg/vaapi/vpp.py
xhaihao: Updated, thx

lib/ffmpeg/vaapi/vpp.py
xhaihao: Updated, thx

uartie (approved):

PR #636, opened 2024-04-08
Add support for FFmpeg-Vaapi drawbox filter

xhaihao:


lib/ffmpeg/vaapi/vpp.py
```diff
@@ -42,6 +42,25 @@ def gen_vpp_opts(self):
...
+      if hasattr(self, "x"):
+        vpopts.append(f"x={self.x}")
+      if hasattr(self, "y"):
+        vpopts.append(f"y={self.y}")
+      if hasattr(self, "color"):
+        vpopts.append(f"color={self.color}")
+      if hasattr(self, "t"):
+        vpopts.append(f"t={self.t}")
+      if hasattr(self, "replace"):
+        vpopts.append(f"t={self.replace}")
```
uartie, line 61: Both use the same ffmpeg option `t`?

lib/mixin/vpp.py
```diff
@@ -224,3 +224,15 @@ def check_color_props(self):
     metrics2.check(**params)
 
   check_range  = check_color_props
+
+  def check_drawbox(self):
+    owidth, oheight = self.width, self.height
```
uartie, line 229: Assignment not needed.  Just pass directly to the `filesize` check

lib/mixin/vpp.py
```diff
@@ -224,3 +224,15 @@ def check_color_props(self):
...
+
+    metrics2.check(
+      metric = dict(type = "filesize"),
+      filetest = self.decoded, width = owidth, height = oheight,
+      frames = self.frames, format = self.format)
+
+    params = vars(self).copy()
+    params.update(width = owidth, height = oheight)
+    metrics2.check(**params)
```
uartie, line 238: Copy/update is unnecessary... nothing changed.  How about call `self.check_default()` instead?

uartie (changes requested):

lib/ffmpeg/vaapi/vpp.py
xhaihao: Thanks for catching this typo, fixed in the new version.

lib/mixin/vpp.py
xhaihao: Thanks for pointing it out, updated.

lib/mixin/vpp.py
xhaihao: Updated, thx

uartie (approved):

PR #639, opened 2024-04-11
[ffmpeg-vaapi]update CQP mode parameters

wenqingx:


lib/ffmpeg/vaapi/encoder.py
```diff
@@ -33,7 +33,7 @@ def profile(self):
   def qp(self):
     def inner(qp):
       if self.codec in [Codec.VP8, Codec.VP9, Codec.AV1]:
-        return f" -global_quality {qp}"
+        return f" -q {qp}"
```
xhaihao, line 36: You may remove this check, and use `-q {qp}` for all codecs, including h264, h265, mpeg2, vp8, vp9, av1.

wenqingx: thx, but I still need some time to verify that this modification will not cause any Regression issues in function/quality/ testing.

lib/ffmpeg/d3d12/encoder.py
wenqingx, line 1: @tong1wu , if have time . please also help review the cmdline.

tong1wu: LGTM

lib/ffmpeg/d3d12/encoder.py
```diff
@@ -32,14 +32,15 @@ def profile(self):
   @property
   def qp(self):
     def inner(qp):
-      if self.codec in [Codec.VP8, Codec.VP9]:
-        return f" -global_quality {qp}"
+      if self.codec in [Codec.AVC, Codec.HEVC, Codec.VP8, Codec.VP9]:
```
xhaihao, line 35: codec.hevc also supports ICQ mode.

lib/ffmpeg/vaapi/encoder.py
```diff
@@ -32,13 +32,11 @@ def profile(self):
   @property
   def qp(self):
     def inner(qp):
-      if self.codec in [Codec.VP8, Codec.VP9, Codec.AV1]:
-        return f" -global_quality {qp}"
-
+      if self.codec not in [Codec.AVC, Codec.HEVC, Codec.VP8, Codec.VP9, Codec.AV1]:
+        return f" -q {qp}"
```
uartie, line 36: @xhaihao What's the difference between `-q` and `-qp` ffmpeg options?  I see `h264_vaapi` and `h265_vaapi` has `-qp` option.

uartie (changes requested):

lib/ffmpeg/vaapi/encoder.py
xhaihao: -qp is a private option in h264_vaapi and h265_vaapi encoders however -q is a common output option of FFmpeg tool. We may use the same -q option for all qsv, vaapi, d3d12 encoders.

uartie: Thanks!

Can you confirm the mapped QP ranges in the tests are still correct?  They still appear different in the test logic for each plugin.  It would be nice if we could move the qp logic to the base class if all plugins were the same.

xhaihao: Ideally we should move the qp logic to the base class because QP range is codec related. However, as you know, we have some specific requirement in qsv.
Note -q is set in the command line, CQP mode is used, so -q can't be used to set quality based parameter in other RC modes.

lib/ffmpeg/d3d12/encoder.py
```diff
@@ -32,14 +32,15 @@ def profile(self):
...
+        return f" -q {qp}"
+      if self.codec in [Codec.HEVC, Codec.AV1]:
         if "ICQ" == self.rcmode:
           return f" -global_quality {qp}"
+        return f" -q {qp}"
       if self.codec in [Codec.MPEG2]:
         mqp = mapRangeInt(qp, [0, 100], [1, 31])
-        return f" -global_quality {mqp}"
+        return f" -q {mqp}"
       return " -qp {qp}"
```
uartie, line 44: Is `-qp` a valid option in d3d12 plugins?  How about we `return f" -q {qp}"` by default, instead?

lib/ffmpeg/d3d12/encoder.py
```diff
@@ -32,14 +32,15 @@ def profile(self):
...
     def inner(qp):
-      if self.codec in [Codec.VP8, Codec.VP9]:
-        return f" -global_quality {qp}"
-      if self.codec in [Codec.AV1]:
+      if self.codec in [Codec.AVC, Codec.VP8, Codec.VP9]:
+        return f" -q {qp}"
+      if self.codec in [Codec.HEVC, Codec.AV1]:
         if "ICQ" == self.rcmode:
           return f" -global_quality {qp}"
+        return f" -q {qp}"
```
uartie, line 40: this can be removed if we make this the default return value

uartie (changes requested):

lib/ffmpeg/d3d12/encoder.py
wenqingx: @tong1wu please help a double check

tong1wu: D3D12 is identical to VAAPI. So just use the same option as vaapi(-q).

tong1wu:
Please change commit log from ffmpeg-d3d11 to ffmpeg-d3d12. Thx.

lib/ffmpeg/vaapi/encoder.py
uartie: much simpler, thx.

wenqingx:
> Please change commit log from ffmpeg-d3d11 to ffmpeg-d3d12. Thx.

updated

uartie (changes requested):
Please rebase to fix pre-check

wenqingx:
> Please rebase to fix pre-check

fixed

PR #642, opened 2024-04-22
caps: update d3d12 encoder caps for DG2

tong1wu:
D3D12 should use encoder instead of vdenc.

### DIFF
--- a/lib/caps/DG2/d3d12
+++ b/lib/caps/DG2/d3d12
@@ -27,17 +27,18 @@ caps = dict(
     av1_8   = dict(maxres = res16k , fmts = ["NV12"]),
     av1_10  = dict(maxres = res16k , fmts = ["P010"]),
   ),
-  vdenc   = dict(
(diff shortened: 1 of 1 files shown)


uartie:
> D3D12 should use encoder instead of vdenc.

All the other DG2 driver caps use vdenc.  So why does D3D12 not?  The hardware is vdenc only, right?  So why doesn't D3D12 make that distinction?

tong1wu:
> > D3D12 should use encoder instead of vdenc.
> 
> All the other DG2 driver caps use vdenc. So why does D3D12 not? The hardware is vdenc only, right? So why doesn't D3D12 make that distinction?

D3D12 has no such low_power option and vdenc concepts. When I initially added the lib/ffmpeg/d3d12/encoder.py, I removed all the LP test suites. And we also use encode caps for vaon12 on DG2. Yes the hardware is vdenc only but from FFmpeg level could we use "encode" caps for simplicity? What do you think Artie? Thank you.

uartie (approved):
Thanks for the explanation

PR #648, opened 2024-05-11
gst-d3d11: update the caps of gst-d3d11

tong1wu:
To align with ffmpeg-d3d11, we remove the 12bit decode and update the supported formats.


lib/gstreamer/d3d11/decoder.py
```diff
@@ -26,6 +26,13 @@ class DecoderTest(BaseDecoderTest):
   def before(self):
     super().before()
 
+  def get_supported_format_map(self):
+    return {
+      "I420"  : "yuv420p",
+      "NV12"  : "nv12",
+      "P010"  : "p010le",
+    }
+
```
uartie, line 35: The gstreamer base decoder class has not been converted to use the BaseFormatMapper, yet.  Only the ffmpeg base decoder class uses it.
Thus, this member function will never be called.

uartie (changes requested):

lib/gstreamer/d3d11/decoder.py
uartie: @FocusLuo can you reassign or close this.

uartie:
no activity

PR #651, opened 2024-05-14
[ffmpeg-qsv] add string api support

FocusLuo:
add string api support for ffmpeg-qsv
1, enable on DG2 fistly
2, the patch comments will update later after some times reviews done

### DIFF
--- a/lib/ffmpeg/qsv/encoder.py
+++ b/lib/ffmpeg/qsv/encoder.py
@@ -15,6 +15,7 @@ from ....lib.ffmpeg.qsv.decoder import Decoder
 from ....lib.common import mapRangeInt, get_media, call, exe2os
 from ....lib.codecs import Codec
 from ....lib.formats import PixelFormat
+from ....lib.mfx.api import StringAPI
 
 class Encoder(FFEncoder):
   hwaccel   = property(lambda s: "qsv")
@@ -64,6 +65,13 @@ class Encoder(FFEncoder):
       f"{self.iqfactor}{self.bqfactor}"
       f"{self.iqoffset}{self.bqoffset}"
     )
+
+# NOTE: Inherit from the StringAPI class first so that it overrides the
+# Encoder class
+class StringAPIEncoder(StringAPI, Encoder):
+   @property
+   def encparams(self):
+     return f" -qsv_params '{super().encparams}'"
 
 @slash.requires(*have_ffmpeg_hwaccel("qsv"))
 @slash.requires(using_compatible_driver)
--- a/lib/ffmpeg/util.py
+++ b/lib/ffmpeg/util.py
@@ -86,6 +86,17 @@ def have_ffmpeg_encoder(encoder):
   return result, encoder
 
 @memoize
+def have_ffmpeg_encoder_options(name, *args):
+  failmsg = "{0}.{1}".format(name, '.'.join(args))
+  result = try_call(f"{exe2os('ffmpeg')} -hide_banner -encoders | awk '{{print $2}}' | grep -w {name}")
+  if result is True:
+    for key in args:
+      result = try_call(f"{exe2os('ffmpeg')} -hide_banner -h encoder={name} | grep -e {key}")
+      if result is False:
+        break
+  return result, failmsg
+
+@memoize
 def have_ffmpeg_decoder(decoder):
   result = try_call(f"{exe2os('ffmpeg')} -hide_banner -decoders | awk '{{print $2}}' | grep -w {decoder}")
   return result, decoder
--- a/lib/mfx/api.py
+++ b/lib/mfx/api.py
@@ -44,76 +44,6 @@ class CodecProfile(IntEnum):
     MFX_PROFILE_VP9_2                       = 3
     MFX_PROFILE_VP9_3                       = 4
 
-class CodecLevel(IntEnum):
-    MFX_LEVEL_UNKNOWN                       =0 #/*!< Unspecified level. */
-    #/* H.264 level 1-1.3 */
-    MFX_LEVEL_AVC_1                         =10
-    MFX_LEVEL_AVC_1b                        =9
-    MFX_LEVEL_AVC_11                        =11
-    MFX_LEVEL_AVC_12                        =12
-    MFX_LEVEL_AVC_13                        =13
-    #/* H.264 level 2-2.2 */
-    MFX_LEVEL_AVC_2                         =20
-    MFX_LEVEL_AVC_21                        =21
-    MFX_LEVEL_AVC_22                        =22
-    #/* H.264 level 3-3.2 */
-    MFX_LEVEL_AVC_3                         =30
-    MFX_LEVEL_AVC_31                        =31
-    MFX_LEVEL_AVC_32                        =32
-    #/* H.264 level 4-4.2 */
-    MFX_LEVEL_AVC_4                         =40
-    MFX_LEVEL_AVC_41                        =41
-    MFX_LEVEL_AVC_42                        =42
-    #/* H.264 level 5-5.2 */
-    MFX_LEVEL_AVC_5                         =50
-    MFX_LEVEL_AVC_51                        =51
-    MFX_LEVEL_AVC_52                        =52
-    #/* H.264 level 6-6.2 */
-    MFX_LEVEL_AVC_6                         =60
-    MFX_LEVEL_AVC_61                        =61
-    MFX_LEVEL_AVC_62                        =62
-
-    #/* HEVC levels */
-    MFX_LEVEL_HEVC_1   = 10
-    MFX_LEVEL_HEVC_2   = 20
-    MFX_LEVEL_HEVC_21  = 21
-    MFX_LEVEL_HEVC_3   = 30
-    MFX_LEVEL_HEVC_31  = 31
-    MFX_LEVEL_HEVC_4   = 40
-    MFX_LEVEL_HEVC_41  = 41
-    MFX_LEVEL_HEVC_5   = 50
-    MFX_LEVEL_HEVC_51  = 51
-    MFX_LEVEL_HEVC_52  = 52
-    MFX_LEVEL_HEVC_6   = 60
-    MFX_LEVEL_HEVC_61  = 61
-    MFX_LEVEL_HEVC_62  = 62
-
-    #/* AV1 Levels */
-    MFX_LEVEL_AV1_2                         = 20
-    MFX_LEVEL_AV1_21                        = 21
-    MFX_LEVEL_AV1_22                        = 22
-    MFX_LEVEL_AV1_23                        = 23
-    MFX_LEVEL_AV1_3                         = 30
-    MFX_LEVEL_AV1_31                        = 31
-    MFX_LEVEL_AV1_32                        = 32
-    MFX_LEVEL_AV1_33                        = 33
-    MFX_LEVEL_AV1_4                         = 40
-    MFX_LEVEL_AV1_41                        = 41
-    MFX_LEVEL_AV1_42                        = 42
-    MFX_LEVEL_AV1_43                        = 43
-    MFX_LEVEL_AV1_5                         = 50
-    MFX_LEVEL_AV1_51                        = 51
-    MFX_LEVEL_AV1_52                        = 52
-    MFX_LEVEL_AV1_53                        = 53
-    MFX_LEVEL_AV1_6                         = 60
-    MFX_LEVEL_AV1_61                        = 61
-    MFX_LEVEL_AV1_62                        = 62
-    MFX_LEVEL_AV1_63                        = 63
-    MFX_LEVEL_AV1_7                         = 70
-    MFX_LEVEL_AV1_71                        = 71
-    MFX_LEVEL_AV1_72                        = 72
-    MFX_LEVEL_AV1_73                        = 73
-
 class RateControlMethod(IntEnum):
 #/*! The RateControlMethod enumerator itemizes bitrate control methods. */
     MFX_RATECONTROL_CBR       =1 #/*!< Use the constant bitrate control algorithm. */
@@ -168,6 +98,7 @@ class RateControlMethod(IntEnum):
     MFX_RATECONTROL_QVBR      =14
 
 class StringAPI(PropertyHandler):
+  level   = property(lambda s: s.ifprop("level", "CodecLevel={level}"))
   gop     = property(lambda s: s.ifprop("gop", "GopPicSize={gop}"))
   bframes = property(lambda s: s.ifprop("bframes", "GopRefDist={bframes}"))
   slices  = property(lambda s: s.ifprop("slices", "NumSlice={slices}"))
@@ -194,13 +125,6 @@ class StringAPI(PropertyHandler):
   @property
   def rcmode(self):
     return f"RateControlMethod={self.map_rcmode(self.props['rcmode'])}"
-
-  @property
-  def level(self):
-    def inner(level):
-      # TODO: need to define map_level
-      return f"CodecLevel={self.map_level(self.props['level'])}"
-    return self.ifprop("level", inner)
 
   @property
   def qp(self):
@@ -259,80 +183,6 @@ class StringAPI(PropertyHandler):
         "profile3"  : CodecProfile.MFX_PROFILE_VP9_3
       },
     }[codec][profile]
-
-  @classmethod
-  @memoize
-  def map_level(cls, codec, level):
-    # TODO: we could move this method to the enum class as "lookup"
-    # then call MfxCodecProfile.lookup(codec, profile)
-    return {
-      #/* H264 levels */
-      Codec.AVC     : {
-        10  : MFX_LEVEL_AV1_1,    
-        9   : MFX_LEVEL_AV1_1b,    
-        11  : MFX_LEVEL_AV1_11,    
-        12  : MFX_LEVEL_AV1_12,    
-        13  : MFX_LEVEL_AV1_13,    
-        20  : MFX_LEVEL_AV1_2,    
-        21  : MFX_LEVEL_AV1_21,    
-        22  : MFX_LEVEL_AV1_22,    
-        30  : MFX_LEVEL_AV1_3,    
-        31  : MFX_LEVEL_AV1_31,    
-        32  : MFX_LEVEL_AV1_32,    
-        40  : MFX_LEVEL_AV1_4,    
-        41  : MFX_LEVEL_AV1_41,    
-        42  : MFX_LEVEL_AV1_42,    
-        50  : MFX_LEVEL_AV1_5,    
-        51  : MFX_LEVEL_AV1_51,    
-        52  : MFX_LEVEL_AV1_52,    
-        60  : MFX_LEVEL_AV1_6,    
-        61  : MFX_LEVEL_AV1_61,    
-        62  : MFX_LEVEL_AV1_62,    
-      },
-      #/* HEVC levels */
-      Codec.HEVC   : {
-        10  : MFX_LEVEL_HEVC_1,    
-        20  : MFX_LEVEL_HEVC_2,    
-        21  : MFX_LEVEL_HEVC_21,    
-        30  : MFX_LEVEL_HEVC_3,    
-        31  : MFX_LEVEL_HEVC_31,    
-        40  : MFX_LEVEL_HEVC_4,    
-        41  : MFX_LEVEL_HEVC_41,    
-        50  : MFX_LEVEL_HEVC_5,    
-        51  : MFX_LEVEL_HEVC_51,    
-        52  : MFX_LEVEL_HEVC_52,    
-        60  : MFX_LEVEL_HEVC_6,    
-        61  : MFX_LEVEL_HEVC_61,    
-        62  : MFX_LEVEL_HEVC_62,    
-      },
-      #/* AV1 Levels */
-      Codec.AV1 : {
-        20  : MFX_LEVEL_AV1_2,    
-        21  : MFX_LEVEL_AV1_21,    
-        22  : MFX_LEVEL_AV1_22,    
-        23  : MFX_LEVEL_AV1_23,    
-        30  : MFX_LEVEL_AV1_3,    
-        31  : MFX_LEVEL_AV1_31,    
-        32  : MFX_LEVEL_AV1_32,    
-        33  : MFX_LEVEL_AV1_33,    
-        40  : MFX_LEVEL_AV1_4,    
-        41  : MFX_LEVEL_AV1_41,    
-        42  : MFX_LEVEL_AV1_42,    
-        43  : MFX_LEVEL_AV1_43,    
-        50  : MFX_LEVEL_AV1_5,    
-        51  : MFX_LEVEL_AV1_51,    
-        52  : MFX_LEVEL_AV1_52,    
-        53  : MFX_LEVEL_AV1_53,    
-        60  : MFX_LEVEL_AV1_6,    
-        61  : MFX_LEVEL_AV1_61,    
-        62  : MFX_LEVEL_AV1_62,    
-        63  : MFX_LEVEL_AV1_63,    
-        70  : MFX_LEVEL_AV1_7,    
-        71  : MFX_LEVEL_AV1_71,    
-        72  : MFX_LEVEL_AV1_72,    
-        73  : MFX_LEVEL_AV1_73,    
-      },
-    }[codec][level]
 
   @classmethod
   @memoize
--- a/lib/mfx/api.py
+++ b/lib/mfx/api.py
@@ -1,0 +1,347 @@
+from ...lib.codecs import Codec
+from ...lib.common import memoize
+from ...lib.properties import PropertyHandler
+from enum import IntEnum, unique
+
+#define come from _install/include/vpl/mfxstructures.h
+class CodecProfile(IntEnum):
+    MFX_PROFILE_UNKNOWN                     =0 #/*!< Unspecified profile. */
+
+    #/* Combined with H.264 profile these flags impose additional constrains. See H.264 specification for the list of constrains. */
+    MFX_PROFILE_AVC_CONSTRAINT_SET0     = (0x100 << 0)
+    MFX_PROFILE_AVC_CONSTRAINT_SET1     = (0x100 << 1)
+    MFX_PROFILE_AVC_CONSTRAINT_SET2     = (0x100 << 2)
+    MFX_PROFILE_AVC_CONSTRAINT_SET3     = (0x100 << 3)
+    MFX_PROFILE_AVC_CONSTRAINT_SET4     = (0x100 << 4)
+    MFX_PROFILE_AVC_CONSTRAINT_SET5     = (0x100 << 5)
+
+    #/* H.264 Profiles. */
+    MFX_PROFILE_AVC_BASELINE                =66
+    MFX_PROFILE_AVC_MAIN                    =77
+    MFX_PROFILE_AVC_EXTENDED                =88
+    MFX_PROFILE_AVC_HIGH                    =100
+    MFX_PROFILE_AVC_HIGH10                  =110
+    MFX_PROFILE_AVC_HIGH_422                =122
+    MFX_PROFILE_AVC_CONSTRAINED_BASELINE    =MFX_PROFILE_AVC_BASELINE + MFX_PROFILE_AVC_CONSTRAINT_SET1
+    MFX_PROFILE_AVC_CONSTRAINED_HIGH        =MFX_PROFILE_AVC_HIGH     + MFX_PROFILE_AVC_CONSTRAINT_SET4 + MFX_PROFILE_AVC_CONSTRAINT_SET5
+    MFX_PROFILE_AVC_PROGRESSIVE_HIGH        =MFX_PROFILE_AVC_HIGH     + MFX_PROFILE_AVC_CONSTRAINT_SET4
+
+    #/* HEVC profiles */
+    MFX_PROFILE_HEVC_MAIN             =1
+    MFX_PROFILE_HEVC_MAIN10           =2
+    MFX_PROFILE_HEVC_MAINSP           =3
+    MFX_PROFILE_HEVC_REXT             =4
+    MFX_PROFILE_HEVC_SCC              =9
+
+    #/* AV1 Profiles */
+    MFX_PROFILE_AV1_MAIN                    = 1
+    MFX_PROFILE_AV1_HIGH                    = 2
+    MFX_PROFILE_AV1_PRO                     = 3
+
+    #/* VP9 Profiles */
+    MFX_PROFILE_VP9_0                       = 1
+    MFX_PROFILE_VP9_1                       = 2
+    MFX_PROFILE_VP9_2                       = 3
+    MFX_PROFILE_VP9_3                       = 4
+
+class CodecLevel(IntEnum):
+    MFX_LEVEL_UNKNOWN                       =0 #/*!< Unspecified level. */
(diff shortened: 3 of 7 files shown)


FocusLuo:
ffmpeg-qsv cmdline:
./smoke run test/ffmpeg-qsv/encode/avc.py --pl DG2 --context=sanity --device=/dev/dri/renderD129 --stringapi 2 -vv
stringapi = 0 and others, to use traditional ffmpeg encode cmdline option
stringapi = 1, to use string api ffmpeg encode cmdline option
stringapi = 2, random to use traditional or string api ffmpeg encode cmdline option

lib/ffmpeg/qsv/encoder.py
```diff
@@ -57,13 +60,165 @@ def inner(quality):
...
+        _encparams = f"{_encparams}:TargetUsage={_quality}"
+    if self.lowpower != None and len(self.lowpower.strip())>0:
+        _lowpower=re.findall("\d+", self.lowpower)[0]
+        _encparams = f"{_encparams}:LowPower={_lowpower}"
+    #{self.forced_idr} ffmpeg
+    #{self.pict}{self.rqp}
+
+    if _encparams != None and len(_encparams.strip()) > 1:
+        if ':' == _encparams[0]:
+            _encparams = _encparams[1:]
```
uartie, line 173: `_encparams` will never be `None`, so this could be `_encparams = _encparams.lstrip(':')`.

That said, how about make _encparams a list instead... append each option as needed, then just do `return ':'.join(_encparams)`

.slashrc
```diff
@@ -88,6 +88,12 @@ class MediaPlugin(slash.plugins.PluginInterface):
     parser.add_argument("--parallel-metrics", action = "store_true")
     parser.add_argument("--platform", default = None, type = str, required = True)
     parser.add_argument("--device", default = '/dev/dri/renderD128', type = str, help = 'vaapi-fits run with render device')
+    parser.add_argument("--stringapi", default = 0,
+      type = int, help = (
+        f"{int(0)}   = default, disabled string api; "
+        f"{int(1)}   = only enabled string api; "
+        f"{int(2)}   = random call parameters with enabling and disabling string api"
+        " (default: %(default)i)"))
```
uartie, line 96: Hmm... this option is specific to ffmpeg-qsv only and I don't like exposing it at the framework core.  We should think of a better way.  Maybe just create string api specific test functions in ffmpeg-qsv??? IDK.

lib/caps/DG2/info
```diff
@@ -10,4 +10,5 @@
 
 info = dict(
   gpu = dict(gen = 12.7),
+  stringapi = dict(enable = True),
```
uartie, line 13: Also seems out of place.  This gives the impression that string api will only work on specific platforms.

uartie (changes requested):
This is not a full review...

I'm not convinced we should modify the base qsv encoder.  It seems we could introduce a new qsv string encoder class or something.  Let me think about the design some more and we can discuss later...

lib/ffmpeg/qsv/encoder.py
FocusLuo: sure, thanks 

> `_encparams` will never be `None`, so this could be `_encparams = _encparams.lstrip(':')`.
> 
> That said, how about make _encparams a list instead... append each option as needed, then just do `return ':'.join(_encparams)`

Sure, thanks suggestion

lib/caps/DG2/info
FocusLuo: > Also seems out of place. This gives the impression that string api will only work on specific platforms.
the stringapi is only enabling on libvpl, so there are platforms limited, so added platform caps define.
do you have any suggestion?

.slashrc
FocusLuo: > Hmm... this option is specific to ffmpeg-qsv only and I don't like exposing it at the framework core. We should think of a better way. Maybe just create string api specific test functions in ffmpeg-qsv??? IDK.
hi Artie, gst-libvpl will enable soon. Mengke is on developping gstreamer source codes.

FocusLuo:
> This is not a full review...
> 
> I'm not convinced we should modify the base qsv encoder. It seems we could introduce a new qsv string encoder class or something. Let me think about the design some more and we can discuss later...

Cool! thanks a lot Artie on supporting this. BTW, gst-vpl will also support stringAPI in Q3, so please also think about gst-vpl part.

uartie:
> > This is not a full review...
> > I'm not convinced we should modify the base qsv encoder. It seems we could introduce a new qsv string encoder class or something. Let me think about the design some more and we can discuss later...
> 
> Cool! thanks a lot Artie on supporting this. BTW, gst-vpl will also support stringAPI in Q3, so please also think about gst-vpl part.

@FocusLuo please have a look at https://github.com/uartie/vaapi-fits/commit/714ea0621a74c1b946e118134a16c3741bf05001
It uses multi-inheritance techniques to "mixin" the string api encoder overrides.  It isn't a complete solution, yet, so I figure you may be able to finish some of the details (e.g. additional property overrides, mappings, etc.).
Note, at the end of the file is a quick example of usage.  We may need to think about if we want to reuse existing user config spec variants or add a new variant.  If we reuse, then we will essentially double the encode cases, which may not be desirable.

.slashrc
uartie: Right.  And I don't think we should do that.  I think they should be new tests that are added to current test context without requiring us to change the CI test command line.

FocusLuo:
> > > This is not a full review...
> > > I'm not convinced we should modify the base qsv encoder. It seems we could introduce a new qsv string encoder class or something. Let me think about the design some more and we can discuss later...
> > 
> > 
> > Cool! thanks a lot Artie on supporting this. BTW, gst-vpl will also support stringAPI in Q3, so please also think about gst-vpl part.
> 
> @FocusLuo please have a look at [uartie@714ea06](https://github.com/uartie/vaapi-fits/commit/714ea0621a74c1b946e118134a16c3741bf05001) It uses multi-inheritance techniques to "mixin" the string api encoder overrides. It isn't a complete solution, yet, so I figure you may be able to finish some of the details (e.g. additional property overrides, mappings, etc.). Note, at the end of the file is a quick example of usage. We may need to think about if we want to reuse existing user config spec variants or add a new variant. If we reuse, then we will essentially double the encode cases, which may not be desirable.

Thanks Artie, by re-used the https://github.com/uartie/vaapi-fits/commit/714ea0621a74c1b946e118134a16c3741bf05001, 
re-submitted the patches, how do you think?
1) only enabled the avc/hevc/av1 encode firstly
2) there will be some failed cases on cbr and vbr, especailly for hevc and av1 encode, how to control the CI regression test report? do you have any idea?

FocusLuo:
New test cases cmdlines: for avc/hevc/av1 encode
 ./smoke run test/ffmpeg-qsv/encode/hevc.py:cqp.test_strapi --pl ADL --context=sanity --device=/dev/dri/renderD128 -vv
 ./smoke run test/ffmpeg-qsv/encode/hevc.py:cqp_lp.test_strapi --pl DG2--context=sanity --device=/dev/dri/renderD129 -vv
 ./smoke run test/ffmpeg-qsv/encode/hevc.py:cbr.test_strapi --pl ADL --context=sanity --device=/dev/dri/renderD128 -vv
 ./smoke run test/ffmpeg-qsv/encode/hevc.py:cbr_lp.test_strapi --pl DG2--context=sanity --device=/dev/dri/renderD129 -vv
./smoke run test/ffmpeg-qsv/encode/hevc.py:vbr.test_strapi --pl ADL --context=sanity --device=/dev/dri/renderD128 -vv
 ./smoke run test/ffmpeg-qsv/encode/hevc.py:vbr_lp.test_strapi --pl DG2--context=sanity --device=/dev/dri/renderD129 -vv

lib/platform.py
```diff
@@ -85,6 +85,13 @@ def have_caps(*args):
   )
   return get_caps(*args) is not None, failmsg
 
+@memoize
+def have_string_api():
+  from .common import get_media
+  has_strapi = False
+  has_strapi = get_media()._get_platform_stringapi_enable()
+  return has_strapi is not None, False
+
```
uartie, line 94: I don't think this should be a platform cap.  The string api is not platform dependent... rather, it is middleware plugin dependent.

I envision this would be defined at the appropriate middleware util file, instead.  For example, if the option is global, in `lib/ffmpeg/util.py` we would add a method to probe for an ffmpeg option (similar to `have_ffmpeg_filter_options`).  If the option is a filter level option, then `have_ffmpeg_filter_options` can be used directly.  Then in `lib/ffmpeg/qsv/util.py` we can define `have_string_api` that simply calls that option probe with `qsv_params` as the parameter.

lib/mfx/api.py
```diff
@@ -0,0 +1,141 @@
+from ...lib.codecs import Codec
+from ...lib.common import memoize
+from ...lib.properties import PropertyHandler
+
+# TODO: move string_api enums to this file
+# NOTE/FIXME: these enums need to derive from IntEnum, not Enum
+from ...lib.string_api import CodecProfile as MfxCodecProfile
+from ...lib.string_api import CodecLevel as MfxCodecLevel
+from ...lib.string_api import RateControlMethod as MfxRateControlMethod
```
uartie, line 9: Please resolve this TODO.

lib/mfx/api.py
```diff
@@ -0,0 +1,141 @@
...
+
+class StringAPI(PropertyHandler):
+  level   = property(lambda s: s.ifprop("level", "CodecLevel={level}"))
```
uartie, line 12: This `level` property definition can be removed as it is defined again, below.

lib/mfx/api.py
```diff
@@ -0,0 +1,141 @@
...
+
+  @property
+  def rcmode(self):
+    return f"RateControlMethod={self.map_rcmode(self.props['rcmode'])}"
+
+  @property
+  def level(self):
+    def inner(level):
+      # TODO: need to define map_level
+      return f"CodecLevel={self.map_level(self.props['level'])}"
```
uartie, line 44: Please resolve this TODO.  The map_level should be implemented in the appropriate subclass.

lib/mfx/api.py
```diff
@@ -0,0 +1,141 @@
...
+      self.extbrc, self.ldb, self.strict, self.pict, self.quality, self.lowpower,
+      self.tilecols, self.tilerows, self.maxframesize,
+    ]
+    return ':'.join(v for v in params if len(v) != 0)
+
+  @classmethod
+  @memoize
+  def map_profile(cls, codec, profile):
+    # TODO: we could move this method to the enum class as "lookup"
+    # then call MfxCodecProfile.lookup(codec, profile)
```
uartie, line 75: Please resolve this TODO.

lib/mfx/api.py
```diff
@@ -0,0 +1,141 @@
...
+        "profile2"  : MfxCodecProfile.MFX_PROFILE_VP9_2,
+        "profile3"  : MfxCodecProfile.MFX_PROFILE_VP9_3
+      },
+    }[codec][profile]
+
+  @classmethod
+  @memoize
+  def map_rcmode(cls, rcmode):
+    # TODO: we could move this method to the enum class as "lookup"
+    # then call MfxRateControlMethod.lookup(rcmode)
```
uartie, line 109: Please resolve this TODO.

test/ffmpeg-qsv/encode/av1.py
```diff
@@ -37,6 +38,13 @@ def test_r2r(self, case, gop, bframes, tilecols, tilerows, qp, quality, profile)
     vars(self).setdefault("r2r", 5)
     self.encode()
 
+  @slash.requires(have_string_api)
+  @slash.parametrize(*gen_av1_cqp_parameters(spec))
```
uartie, line 42: Upon second thought, I don't think we should reuse these gen functions because we'll essentially double the number of cases that are already defined for this config variant.  We should introduce a new variant ("strapi"?) in the config files.  This variant could be generalized to include the rcmode parameter similar to the `pict` tests so we don't have to write a test function for every rcmode test class.

uartie (changes requested):

uartie:
> > > > This is not a full review...
> > > > I'm not convinced we should modify the base qsv encoder. It seems we could introduce a new qsv string encoder class or something. Let me think about the design some more and we can discuss later...
> > > 
> > > 
> > > Cool! thanks a lot Artie on supporting this. BTW, gst-vpl will also support stringAPI in Q3, so please also think about gst-vpl part.
> > 
> > 
> > @FocusLuo please have a look at [uartie@714ea06](https://github.com/uartie/vaapi-fits/commit/714ea0621a74c1b946e118134a16c3741bf05001) It uses multi-inheritance techniques to "mixin" the string api encoder overrides. It isn't a complete solution, yet, so I figure you may be able to finish some of the details (e.g. additional property overrides, mappings, etc.). Note, at the end of the file is a quick example of usage. We may need to think about if we want to reuse existing user config spec variants or add a new variant. If we reuse, then we will essentially double the encode cases, which may not be desirable.
> 
> Thanks Artie, by re-used the [uartie@714ea06](https://github.com/uartie/vaapi-fits/commit/714ea0621a74c1b946e118134a16c3741bf05001), re-submitted the patches, how do you think?
> 
> 1. only enabled the avc/hevc/av1 encode firstly
> 2. there will be some failed cases on cbr and vbr, especailly for hevc and av1 encode, how to control the CI regression test report? do you have any idea?

My patch wasn't a completed solution... please finish the TODO's and other details (missing some property definitions and mappings).

lib/mfx/api.py
FocusLuo: the next patch will move map_rate  to the enum class as lookup

lib/mfx/api.py
FocusLuo: the next patch will move map_profile  to the enum class as lookup

test/ffmpeg-qsv/encode/av1.py
FocusLuo: ok, let me to update and submit another version

FocusLuo:
@uartie do you think it is good to merge this PR firstly? then let me to submit another PR for your other comments such as 
1) move map_level/rate/profile to the enum class as lookup; 
2) to use new gen functions for string (there will have avc gen cqp/cbr/vbr, hevc gen functions, av1 gen functions and also have lp and non-lp),
and definitely you will have other comments input in future.
how do you think?  this will help me decrease my local update/comment maintain effort, and also help wenqing to do some issue testing on finding issue among different platforms or integrate into CI issue firstly
thanks
Focus

uartie:
> @uartie do you think it is good to merge this PR firstly? then let me to submit another PR for your other comments such as
> 
> 1. move map_level/rate/profile to the enum class as lookup;
> 2. to use new gen functions for string (there will have avc gen cqp/cbr/vbr, hevc gen functions, av1 gen functions and also have lp and non-lp),
>    and definitely you will have other comments input in future.
>    how do you think?  this will help me decrease my local update/comment maintain effort, and also help wenqing to do some issue testing on finding issue among different platforms or integrate into CI issue firstly
>    thanks
>    Focus

The issue I have is it automatically doubles the cases as mentioned in https://github.com/intel/vaapi-fits/pull/651#discussion_r1644879125.  For full tests, it would add an additional 2-3 hours depending on platform.

FocusLuo:
> > @uartie do you think it is good to merge this PR firstly? then let me to submit another PR for your other comments such as
> > 
> > 1. move map_level/rate/profile to the enum class as lookup;
> > 2. to use new gen functions for string (there will have avc gen cqp/cbr/vbr, hevc gen functions, av1 gen functions and also have lp and non-lp),
> >    and definitely you will have other comments input in future.
> >    how do you think?  this will help me decrease my local update/comment maintain effort, and also help wenqing to do some issue testing on finding issue among different platforms or integrate into CI issue firstly
> >    thanks
> >    Focus
> 
> The issue I have is it automatically doubles the cases as mentioned in [#651 (comment)](https://github.com/intel/vaapi-fits/pull/651#discussion_r1644879125). For full tests, it would add an additional 2-3 hours depending on platform.

Ok, will submit new patch to use dedicated gen functions for string api, thanks

test/ffmpeg-qsv/encode/av1.py
FocusLuo: updated, used dedicated parameter for str api
and firstly added avc/hevc 8bit/av1 encode cqp/cbr/vbr with vme and lp supported firstly

lib/parameters_strapi.py
```diff
@@ -0,0 +1,273 @@
+###
+### Copyright (C) 2024 Intel Corporation
+###
+### SPDX-License-Identifier: BSD-3-Clause
+###
+
+import copy
+import itertools
+
+def gen_avc_cqp_strapi_variants(spec, profiles):
```
uartie, line 10: I think I pointed it out before...
But, how about we consolidate the parameters from all rcmode into a single generator? It would look similar to `gen_avc_forced_idr_variants`.  It would reduce a lot of duplication and lines of code.  So, perhaps the test signature/keys for all avc and hevc could be `("case", "rcmode", "gop", "slices", "bframes", "bitrate", "qp", "quality", "fps", "lowpower", "profile")` where their values are `None` when they don't make sense for the given rcmode.  For av1, remove slices and add tilecols and tilerows?

lib/parameters_strapi.py
```diff
@@ -0,0 +1,273 @@
...
+      # backwards compatibility for ipbmode user config
+      ipbmode = variant.get("ipbmode", None)
+      if ipbmode is not None:
+        assert not variant.get("gop", None)
+        assert not variant.get("slices", None)
+        assert not variant.get("bframes", None)
+        variant.update(
+          slices = 4 if ipbmode == 2 else 1,
+          gop = 30 if ipbmode != 0 else 1,
+          bframes = 2 if ipbmode == 2 else 0)
```
uartie, line 33: This is unnecessary since it is a new variant and therefore has no backwards compat issue.

lib/parameters_strapi.py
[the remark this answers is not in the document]
uartie: agree, but with consolidated strapi generators.

lib/parameters_strapi.py
```diff
@@ -0,0 +1,273 @@
...
+###
+### SPDX-License-Identifier: BSD-3-Clause
+###
+
+import copy
+import itertools
+
+def gen_avc_cqp_strapi_variants(spec, profiles):
+  for case, params in spec.items():
+    variants = copy.deepcopy(params.get("variants", dict()).get("cqp_strapi", None))
```
uartie, line 12: If we consolidate the parameters as described, then it would only be "strapi" variant.  We could also pass the "variant" as a parameter to the generator... then it could be reused for other variants as needed.

uartie (changes requested):

lib/ffmpeg/util.py
```diff
@@ -85,6 +85,11 @@ def have_ffmpeg_encoder(encoder):
   result = try_call(f"{exe2os('ffmpeg')} -hide_banner -encoders | awk '{{print $2}}' | grep -w {encoder}")
   return result, encoder
 
+@memoize
+def have_string_api(encoder):
```
uartie, line 89: How about make this more generic?  Call it `have_ffmpeg_encoder_options`, similar to `have_ffmpeg_filter_options`.

.slashrc
```diff
@@ -209,6 +209,10 @@ class MediaPlugin(slash.plugins.PluginInterface):
     from lib.platform import info
     return info()["gpu"]["gen"]
 
+  def _get_platform_stringapi_enable(self):
+    from lib.platform import info
+    return info()["stringapi"]["enable"]
+
```
uartie, line 215: This is not used anymore. Remove.

uartie (changes requested):

lib/parameters_strapi.py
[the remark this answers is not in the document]
FocusLuo: fixed and updated

lib/parameters_strapi.py
FocusLuo: used the consolidated generators, so there is no this issue

lib/parameters_strapi.py
FocusLuo: used the consolidated generators for strapi parameters, so there is no this issue now

.slashrc
FocusLuo: removed

lib/parameters_strapi.py
FocusLuo: updated, thanks suggestion

lib/ffmpeg/util.py
```diff
@@ -85,6 +85,17 @@ def have_ffmpeg_encoder(encoder):
...
+@memoize
+def have_ffmpeg_encoder_options(name, *args):
+  failmsg = "{0}.{1}".format(name, '.'.join(args))
+  result = try_call(f"{exe2os('ffmpeg')} -hide_banner -encoders | awk '{{print $2}}' | grep -w {name}")
+  if result is True:
+    for key in args:
+      result = try_call(f"{exe2os('ffmpeg')} -hide_banner -h encoder={name} | grep -e {key}")
+      if result is False:
+        break
+  return result, failmsg
```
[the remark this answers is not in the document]
FocusLuo: just follow the have_ffmpeg_filter_options define to return the msg
@memoize
def have_ffmpeg_filter_options(name, *args):
  failmsg = "{0}.{1}".format(name, '.'.join(args))
  result = try_call(f"{exe2os('ffmpeg')} -hide_banner -filters | awk '{{print $2}}' | grep -w {name}")
  if result is True:
    for key in args:
      result = try_call(f"{exe2os('ffmpeg')} -hide_banner -h filter={name} | grep -e '^   {key} '")
      if result is False:
        break
  return result, failmsg

FocusLuo:
@uartie is the latest new update version good to you?

lib/parameters.py
```diff
@@ -102,9 +117,20 @@ def gen_hevc_cbr_level_parameters( spec, profiles):
...
+    if strapi is None:
+      variants = copy.deepcopy(params.get("variants", dict()).get("vbr", None))
+    else:
+      variants = copy.deepcopy(params.get("variants", dict()).get("vbr_strapi", None))
+
+    if variants is None:
+      keys = ["gop", "slices", "bframes", "bitrate", "fps", "quality", "refs", "profile"]
+      product  = list(itertools.product([30], [1], [0], [5000], [30], [4], [1], profiles)) # IP, single-slice
+      product += list(itertools.product([30], [4], [2], [5000], [30], [4], [1], profiles)) # IPB, multi-slice
+      variants = [dict(zip(keys, vals)) for vals in product]
```
uartie, line 131: Please don't add this automatic default case variant generator.

uartie (changes requested):

lib/parameters.py
```diff
@@ -114,14 +140,19 @@ def gen_avc_vbr_variants(spec, profiles):
           variant.get("refs", 1), profile
         ]
 
-def gen_avc_vbr_parameters(spec, profiles):
+def gen_avc_vbr_parameters(spec, profiles, strapi=None):
```
uartie, line 143:
```suggestion
def gen_avc_vbr_parameters(spec, profiles, strapi=False):
```
Please change all of these.  Thanks.

uartie (changes requested):

test/ffmpeg-qsv/encode/av1.py
```diff
@@ -37,6 +38,13 @@ def test_r2r(self, case, gop, bframes, tilecols, tilerows, qp, quality, profile)
     vars(self).setdefault("r2r", 5)
     self.encode()
 
+  @slash.requires(*have_ffmpeg_encoder_options("av1_qsv", "qsv_params"))
+  @slash.parametrize(*gen_av1_cqp_strapi_parameters(spec, strapi=True))
```
uartie, line 42: The generator function name is wrong.

lib/parameters.py
```diff
@@ -22,71 +22,82 @@ def format_value(value, **params):
...
-
-      for profile in cprofiles:
-        yield [
-          case, variant["gop"], variant["slices"], variant["bframes"],
-          variant["qp"], variant["quality"], profile
-        ]
-
-def gen_avc_cqp_parameters(spec, profiles):
+    if strapi is True:
+      variants = copy.deepcopy(params.get("variants", dict()).get("cqp_strapi", None))
```
uartie, line 28:
```suggestion
      variants = copy.deepcopy(params.get("variants", dict()).get("cqp_strapi", []))
```

lib/parameters.py
```diff
@@ -22,71 +22,82 @@ def format_value(value, **params):
...
+      variants = copy.deepcopy(params.get("variants", dict()).get("cqp", None))
+    
+      if variants is None:
+        keys = ["gop", "slices", "bframes", "qp", "quality", "profile"]
+        product  = list(itertools.product([1], [1], [0], [14, 28], [1, 4, 7], profiles))  # I, single-slice
+        product += list(itertools.product([30], [1], [0], [14, 28], [1, 4, 7], profiles)) # IP, single-slice
+        product += list(itertools.product([30], [4], [2], [14, 28], [1, 4, 7], profiles)) # IPB, multi-slice
+        variants = [dict(zip(keys, vals)) for vals in product]
+
+    if variants is not None:
```
uartie, line 39:
```suggestion

```

lib/parameters.py
```diff
@@ -22,71 +22,82 @@ def format_value(value, **params):
...
-        yield [
-          case, variant["gop"], variant["slices"], variant["bframes"],
-          variant["bitrate"], variant.get("fps", 30), profile
-        ]
-
-def gen_avc_cbr_parameters(spec, profiles):
+    if strapi is True:
+      variants = copy.deepcopy(params.get("variants", dict()).get("cbr_strapi", None))
+    else:
+      variants = copy.deepcopy(params.get("variants", dict()).get("cbr", None))
```
uartie, line 71:
```suggestion
      variants = copy.deepcopy(params.get("variants", dict()).get("cbr_strapi", []))
    else:
      variants = copy.deepcopy(params.get("variants", dict()).get("cbr", []))
```

lib/parameters.py
```diff
@@ -22,71 +22,82 @@ def format_value(value, **params):
...
+   
+    if variants is not None:
```
uartie, line 73:
```suggestion

```

lib/parameters.py
```diff
@@ -102,26 +113,37 @@ def gen_hevc_cbr_level_parameters( spec, profiles):
...
-          case, variant["gop"], variant["slices"], variant["bframes"],
-          variant["bitrate"], variant.get("fps", 30), variant.get("quality", 4),
-          variant.get("refs", 1), profile
-        ]
-
-def gen_avc_vbr_parameters(spec, profiles):
+    if strapi is True:
+      variants = copy.deepcopy(params.get("variants", dict()).get("vbr_strapi", None))
+    else:
+      variants = copy.deepcopy(params.get("variants", dict()).get("vbr", None))
```
uartie, line 121:
```suggestion
      variants = copy.deepcopy(params.get("variants", dict()).get("vbr_strapi", []))
    else:
      variants = copy.deepcopy(params.get("variants", dict()).get("vbr", []))
```

lib/parameters.py
```diff
@@ -102,26 +113,37 @@ def gen_hevc_cbr_level_parameters( spec, profiles):
...
+
+    if variants is not None:
```
uartie, line 123:
```suggestion

```

lib/parameters.py
```diff
@@ -22,71 +22,82 @@ def format_value(value, **params):
...
-          bframes = 2 if ipbmode == 2 else 0)
-
-      for profile in cprofiles:
-        yield [
-          case, variant["gop"], variant["slices"], variant["bframes"],
-          variant["qp"], variant["quality"], profile
-        ]
-
-def gen_avc_cqp_parameters(spec, profiles):
+    if strapi is True:
```
uartie, line 27:
```suggestion
    if strapi:
```

uartie (changes requested):
Don't use `is True`.

The `variants` should always be a list before hitting the loop.  This will cut down on the diff noise.

lib/parameters.py
```diff
@@ -22,15 +22,19 @@ def format_value(value, **params):
...
+      variants = copy.deepcopy(params.get("variants", dict()).get("cqp_strapi", []))
+    else:
+      variants = copy.deepcopy(params.get("variants", dict()).get("cqp", []))
+    
+      if variants is None:
+        keys = ["gop", "slices", "bframes", "qp", "quality", "profile"]
+        product  = list(itertools.product([1], [1], [0], [14, 28], [1, 4, 7], profiles))  # I, single-slice
+        product += list(itertools.product([30], [1], [0], [14, 28], [1, 4, 7], profiles)) # IP, single-slice
+        product += list(itertools.product([30], [4], [2], [14, 28], [1, 4, 7], profiles)) # IPB, multi-slice
+        variants = [dict(zip(keys, vals)) for vals in product]
```
uartie, line 37: I'm not sure why you changed this one.  This one has to stay `None` by default to avoid changing legacy functionality.  It is the only legacy situation where we auto-generate cases if user config does not define any cases for the `cqp` variant.  

```suggestion
      variants = copy.deepcopy(params.get("variants", dict()).get("cqp", None))
    
      if variants is None:
        keys = ["gop", "slices", "bframes", "qp", "quality", "profile"]
        product  = list(itertools.product([1], [1], [0], [14, 28], [1, 4, 7], profiles))  # I, single-slice
        product += list(itertools.product([30], [1], [0], [14, 28], [1, 4, 7], profiles)) # IP, single-slice
        product += list(itertools.product([30], [4], [2], [14, 28], [1, 4, 7], profiles)) # IPB, multi-slice
        variants = [dict(zip(keys, vals)) for vals in product]
```

lib/mfx/api.py
```diff
@@ -0,0 +1,373 @@
...
+# class cqp(HEVC8EncoderTest):
+#   ...
+#   @slash.requires(*have_string_api) # TODO: define have_string_api
+#   @slash.parametrize(*gen_hevc_cqp_parameters(spec, ['main']))
+#   def test_strapi(self, case, gop, slices, bframes, qp, quality, profile):
+#     self.EncoderClass = StringAPIEncoder
+#     self.init(spec, case, gop, slices, bframes, qp, quality, profile)
+#     self.encode()
+#
+#
```
uartie, line 373: This comment block can be removed.  It is stale.

The actual implementations provided by your PR can be used as a future reference, instead.

uartie (changes requested):
Please avoid making extra changes that weren't suggested and we could be good to go.  Almost there ;)

uartie (changes requested):
The end result looks good to me.

However, please squash your commits into a few logically organized commits, as appropriate.  There should not be any intermediate commits that only address the reviews.  Those changes should be applied to the originals.

FocusLuo:
> The end result looks good to me.
> 
> However, please squash your commits into a few logically organized commits, as appropriate. There should not be any intermediate commits that only address the reviews. Those changes should be applied to the originals.

thanks Artie, 
updated

uartie (changes requested):
Ugh... sorry I didn't notice these before...

lib/mfx/api.py
```diff
@@ -0,0 +1,347 @@
...
+
+  @property
+  def rcmode(self):
+    return f"RateControlMethod={self.map_rcmode(self.props['rcmode'])}"
+
+  @property
+  def level(self):
+    def inner(level):
+      # TODO: need to define map_level
+      return f"CodecLevel={self.map_level(self.props['level'])}"
```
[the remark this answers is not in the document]
uartie: Thus, it can be simplified to:
```python
level = property(lambda s: s.ifprop("level", "CodecLevel={level}"))
```

uartie (approved):

PR #661, opened 2024-05-22
ffmpeg/vaapi/encode: skip tilecols/rows < 1

uartie:
...so user config does not need to workaround it with a post_validate callback.

cc: @wenqingx


wenqingx:
> ...so user config does not need to workaround it with a post_validate callback.
> 
> cc: @wenqingx



> ...so user config does not need to workaround it with a post_validate callback.
> 
> cc: @wenqingx

thx,  also cc: @feiwan1

wenqingx:
> > ...so user config does not need to workaround it with a post_validate callback.
> > cc: @wenqingx
> 
> thx, also cc: @feiwan1

PR #673, opened 2024-07-30
[ffmpeg-vaapi] add b_depth test options for h264 encode

Hangjie22Coder:


lib/parameters.py
```diff
@@ -50,11 +50,11 @@ def gen_avc_cqp_variants(spec, profiles):
       for profile in cprofiles:
         yield [
           case, variant["gop"], variant["slices"], variant["bframes"],
-          variant["qp"], variant["quality"], profile
+          variant["qp"], variant["quality"], profile, variant.get("bdepth", 0)
```
uartie, line 53: ffmpeg-vaapi default bdepth is 1.  Therefore, this could change the test behavior for existing use config cases that don't have bdepth defined.

uartie (changes requested):
We can't just add a new parameter to an existing test.  The test case name is derived from the test function signature and is tracked in history by that name.  If you add a new parameter, then the test case names change and we lose history.

Also, every middleware test of the same name must use the same test function signature.  That is, every test that uses this parameter generator must match their function signature that takes the parameters.  Otherwise, they will be broken.

uartie:
Maybe better to just create a new test function (e.g. `test_bdepth`)

test/ffmpeg-vaapi/encode/avc.py
```diff
@@ -61,6 +61,46 @@ def test_r2r(self, case, gop, slices, bframes, qp, quality, profile):
...
+      profile   = profile,
+      qp        = qp,
+      quality   = quality,
+      rcmode    = "cqp",
+      slices    = slices,
+      bdepth    = bdepth,
+    )
+
+  @slash.parametrize(*gen_avc_bdepth_parameters(spec, ['high', 'main']))
+  def test(self, case, gop, slices, bframes, qp, quality, profile, bdepth):
```
uartie, line 80: How about `(self, case, gop, bframes, bitrate, qp, maxrate, profile, rcmode, bdepth)` instead.  This would allow us to use the same class for different rcmode.  And it would reduce the need for a bunch of specialized generators.
See the `roi` or `intref` test classes for reference.  It would be similar to those (specialized to bdepth cases, of course).

lib/parameters.py
```diff
@@ -58,6 +58,78 @@ def gen_avc_cqp_parameters(spec, profiles):
   params = gen_avc_cqp_variants(spec, profiles)
   return keys, params
 
+def gen_avc_bdepth_variants(spec, profiles):
+  for case, params in spec.items():
+    variants = copy.deepcopy(params.get("variants", dict()).get("cqp", None))
```
uartie, line 63: We shouldn't reuse the "cqp" variant key because test configs already define cases for this. If we reuse it, then we suddenly double the number of "cqp" test cases.
Define a new key, instead (e.g. bdepth).

uartie (changes requested):

test/ffmpeg-vaapi/encode/avc.py
```diff
@@ -212,3 +212,43 @@ def init(self, tspec, case, gop, bframes, bitrate, maxrate, profile, rcmode):
...
+class bdepth(AVCEncoderTest):
+  def init(self, tspec, case, gop, bframes, bitrate, qp, maxrate, profile, rcmode, bdepth):
+    vars(self).update(tspec[case].copy())
+    vars(self).update(
+      rcmode    = rcmode,
+      bframes   = bframes,
+      bitrate   = bitrate,
+      qp        = qp,
+      case      = case,
+      maxrate   = maxrate,
```
uartie, line 225:
```suggestion
      maxrate   = maxrate,
      minrate   = bitrate,
```

test/ffmpeg-vaapi/encode/avc.py
```diff
@@ -212,3 +212,43 @@ def init(self, tspec, case, gop, bframes, bitrate, maxrate, profile, rcmode):
...
+class bdepth_lp(AVCEncoderLPTest):
+  def init(self, tspec, case, gop, bframes, bitrate, qp, maxrate, profile, rcmode, bdepth):
+    vars(self).update(tspec[case].copy())
+    vars(self).update(
+      rcmode    = rcmode,
+      bframes   = bframes,
+      bitrate   = bitrate,
+      qp        = qp,
+      case      = case,
+      maxrate   = maxrate,
```
uartie, line 245:
```suggestion
      maxrate   = maxrate,
      minrate   = bitrate,
```

uartie (changes requested):

uartie:
Also, please fix conflicts

Hangjie22Coder:
> Also, please fix conflicts

fixed

uartie (approved):

PR #662, opened 2024-05-23
Add gstreamer transcode av1 support

wenqingx:
gst-msdk transcode av1
gst-va transcode av1

PR #675, opened 2024-07-31
gst/xcode: fix AV1 demux assumption and add gst-vaapi AV1 xcode test

uartie:
* gst/xcode: choose decode demuxer based on file ext

    In https://github.com/intel/vaapi-fits/pull/662 we made
    the assumption that the user test config will always use
    a webm (matroska) formatted input source file for AV1
    xcode.

    However, our internal AV1 test cases use ivf input
    source files.  Unfortunately, gstreamer does not have
    a ivf muxer for the encode side of the pipeline. Nor is
    it able to discover an elementary AV1 (or VP9) file.
    Thus, when AV1 (or VP9) is the target output format,
    we still need to use the matroska container format
    (matroskamux).

    Thus, dynamically choose the proper demuxer to decode
    the input source file and the output result files
    based on the filename extension.

* gst-vaapi/xcode: add AV1 xcode support

cc: @wenqingx

### DIFF
--- a/lib/gstreamer/msdk/transcoder.py
+++ b/lib/gstreamer/msdk/transcoder.py
@@ -54,10 +54,10 @@ class TranscoderTest(BaseTranscoderTest):
         ),
       ),
       Codec.VP9 : dict(
-        hw = (platform.get_caps("decode", "vp9_8"), have_gst_element(f"msdkvp9dec"), f"matroskademux ! vp9parse ! msdkvp9dec"),
+        hw = (platform.get_caps("decode", "vp9_8"), have_gst_element(f"msdkvp9dec"), "vp9parse ! msdkvp9dec"),
       ),
       Codec.AV1 : dict(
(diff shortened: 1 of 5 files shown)


wenqingx:
> * gst/xcode: choose decode demuxer based on file ext
>   In [Add gstreamer transcode av1 support #662](https://github.com/intel/vaapi-fits/pull/662) we made
>   the assumption that the user test config will always use
>   a webm (matroska) formatted input source file for AV1
>   xcode.
>   However, our internal AV1 test cases use ivf input
>   source files.  Unfortunately, gstreamer does not have
>   a ivf muxer for the encode side of the pipeline. Nor is
>   it able to discover an elementary AV1 (or VP9) file.
>   Thus, when AV1 (or VP9) is the target output format,
>   we still need to use the matroska container format
>   (matroskamux).
>   Thus, dynamically choose the proper demuxer to decode
>   the input source file and the output result files
>   based on the filename extension.
> * gst-vaapi/xcode: add AV1 xcode support
> 
> cc: @wenqingx

I understand, thank you for helping to fix it

PR #679, opened 2024-08-29
[ffmpeg-qsv] add vp9e playback_seek test case

Hangjie22Coder:

PR #685, opened 2024-09-06
[ffmpeg-qsv] add avce playback_seek test case

Hangjie22Coder:


lib/parameters.py
```diff
@@ -469,6 +469,56 @@ def gen_avc_rqp_parameters(spec, profiles):
...
+        yield [
+            case, rcmode, bitrate, variant.get("maxrate", None),
+            variant.get("fps", 25), variant.get("seek", 1)
+            ]
+
+def gen_avc_seek_lp_parameters(spec):
+  keys = ("case", "rcmode", "bitrate", "maxrate", "fps", "seek")
+  params = gen_avc_seek_lp_variants(spec)
+  return keys, params
+
```
uartie, line 521: #679 merged.  Thus, we don't need the duplicate code. Instead, lets do:
```python
gen_avc_seek_parameters = gen_vp9_seek_parameters
gen_avc_seek_lp_parameters = gen_vp9_seek_lp_parameters
```

---
As a side note, in the future it would be nice if we can refactor this file to consolidate all the other duplicate code.  There is a lot of it.  But, it doesn't need to be done for this PR.

test/ffmpeg-qsv/encode/avc.py
```diff
@@ -394,3 +394,39 @@ def init(self, tspec, case, gop, bframes, bitrate, maxrate, profile, rcmode, max
...
+      fps       = fps,
+      seek      = seek,
+    )
+
+  @slash.parametrize(*gen_avc_seek_parameters(spec))
+  def test(self, case, rcmode, bitrate, maxrate, fps, seek):
+    self.init(spec, case, rcmode, bitrate, maxrate, fps, seek)
+    self.encode()
+
+class seek_lp(AVCEncoderTest):
```
uartie, line 416:
```suggestion
class seek_lp(AVCEncoderLPTest):
```

uartie (changes requested):

Hangjie22Coder:
Thanks for the feedback. I will refactor this file in the future to consolidate all the other duplicate code when I have time.
Additionally, there are some issues with seek_lp(AVCEncoderLPTest), and I will submit a PR focusing on that.

uartie (approved):

PR #686, opened 2024-09-09
Revert "Support experimental VVC decoder by adding -strict -2 in FFmp…

feiwan1:
…eg commands"

This reverts commit 58345448eefb41b36a989fc4e01e3a566b9bc354.

VVC native decoder removed experimental flag in master branch, so this commit is unnecessary.

### DIFF
--- a/lib/ffmpeg/decoderbase.py
+++ b/lib/ffmpeg/decoderbase.py
@@ -27,7 +27,6 @@ class Decoder(PropertyHandler, BaseFormatMapper):
   decoded   = property(lambda s: s._decoded)
   osdecoded = property(lambda s: filepath2os(s.decoded))
   hwaccel   = property(lambda s: s.props["hwaccel"])
-  strict    = property(lambda s: s.ifprop("strict", "-strict {strict}"))
 
   width       = property(lambda s: s.props["width"])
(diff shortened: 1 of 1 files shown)


Hangjie22Coder:
LGTM, By the way, I have submitted a PR to remove the 'strict' property for 8-bit/10-bit VVC decode in ffmpeg-vaapi.
See https://github.com/intel/vaapi-fits/pull/687

lib/ffmpeg/decoderbase.py
```diff
@@ -27,7 +27,6 @@ class Decoder(PropertyHandler, BaseFormatMapper):
   decoded   = property(lambda s: s._decoded)
   osdecoded = property(lambda s: filepath2os(s.decoded))
   hwaccel   = property(lambda s: s.props["hwaccel"])
-  strict    = property(lambda s: s.ifprop("strict", "-strict {strict}"))
```
uartie, line 30: We don't need to remove this property from the decoder.  We might need to use it again in the future for something else.  If it is not set by the "test", then it just resolves to an empty string (therefore, not added to command-line).  #687 removes this from the "test" settings... that is enough to remove the flag from the command-line.

PR #689, opened 2024-09-13
[caps] enable ADL/RPL windows support

wenqingx:
copy from caps of TGL, please help double check.
@uartie @FocusLuo @feiwan1


wenqingx:
tested and shared the results , passed.

PR #690, opened 2024-09-23
[ffmpeg-qsv] add jpeg encode playback_seek test case

Hangjie22Coder:


test/ffmpeg-qsv/encode/jpeg.py
```diff
@@ -31,3 +31,21 @@ def test_r2r(self, case, quality):
...
+      bitrate   = bitrate,
+      maxrate   = maxrate,
+      minrate   = bitrate,
+      rcmode    = rcmode,
+      fps       = fps,
+      seek      = seek,
+    )
+
+  @slash.parametrize(*gen_jpeg_seek_parameters(spec))
+  def test(self, case, rcmode, bitrate, maxrate, fps, seek):
```
uartie, line 49: The bitrate and maxrate don't make sense for jpeg.
Also, only cqp mode for jpeg.  So rcmode param is not needed, either.
Hence, jpeg will need a special param generator.

uartie (changes requested):

test/ffmpeg-qsv/encode/jpeg.py
Hangjie22Coder: Thank you for your review and suggestions. I have updated the code by removing the `rcmode`, `bitrate`, and `maxrate `elements as per your recommendation. The modified implementation now focuses only on the necessary parameters (`fps`, and `seek`). Please let me know if further adjustments are needed.

FocusLuo (approved):
LGTM

FocusLuo:
@Hangjie22Coder  could you help to fix the conflict?

uartie (approved):

PR #691, opened 2024-09-23
Fix Input-Output Frame Mismatch in FFmpeg-VAAPI VPP Deinterlace

Hangjie22Coder:


lib/ffmpeg/vppbase.py
```diff
@@ -51,7 +51,7 @@ def gen_output_opts(self):
     opts = "-filter_complex" if self.vpp_op in fcomplex else "-vf"
     opts += f" '{','.join(vpfilter)}'"
     opts += " -pix_fmt {mformat}" if self.vpp_op not in ["csc", "tonemap", "overlay"] else ""
-    opts += " -f rawvideo -fps_mode passthrough -an -vframes {frames} -y {osdecoded}"
+    opts += " -f rawvideo -fps_mode cfr -an -vframes {frames} -y {osdecoded}"
```
uartie, line 54: Can you explain what "cfr" does and why it fixes this?  Also, this will change the command line for all VPP tests... have you verified they don't regress?

uartie (changes requested):

lib/ffmpeg/vppbase.py
Hangjie22Coder: cfr (Constant Frame Rate) forces FFmpeg to maintain a fixed output frame rate, even if the input timestamps are irregular or the frame rate is variable. This ensures that the output frames are evenly spaced in time, and if there are any gaps in the input frames or missing frames due to timestamp issues, FFmpeg will either drop or duplicate frames to keep the frame rate consistent.

Hangjie22Coder: In our tests, when using fps_mode passthrough, FFmpeg outputs frames directly based on the timestamp of the input stream, which can cause some frames to be lost due to discontinuous timestamps or insufficient reference frames. When using CFR mode, FFmpeg avoids frame drops by normalizing timestamps to ensure that each output frame has a fixed time interval. Therefore, this modification can ensure the stability and integrity of the output frame.

Hangjie22Coder: Regarding your concern about the impact on FFmpeg VPP tests, I understand that changing to cfr affects all tests. I have conducted thorough testing across different cases. While the issue of output frame mismatches has been resolved, I noticed that in some cases, there are issues with bitrate_gap and ssim. This indicates that certain test cases are indeed affected by the change.

Hangjie22Coder: I will explore further to see if there’s a better solution to resolve the output frame mismatch issue.

PR #697, opened 2024-10-24
Enhance PSNR Check for Luma (Y) Channel in VPP Sharpen Filter

Hangjie22Coder:
The current implementation of the PSNR check for the Luma (Y) channel raises exceptions for actual values exceeding reference values. This behavior is not reflective of typical processing scenarios where higher actual values are expected.

Updated the `compare` function to only raise exceptions if the actual Luma (Y) value is lower than the reference value.


Hangjie22Coder:
@Bin-CI @uartie Please help a reivew.

Hangjie22Coder:
Hi @feiwan1 , can you review this for me?

FocusLuo (approved):
LGTM

wenqingx:
I think it 's difficult to detect the situation of rising in a timely manner and it is easy to miss the situation of rising and then falling again.

wenqingx:
LGTM

lib/mixin/vpp.py
```diff
@@ -132,7 +132,10 @@ def compare(k, ref, actual):
         assert actual[-2] == 100, "Cb(U) should not be affected by SHARPEN filter"
         assert actual[-1] == 100, "Cr(V) should not be affected by SHARPEN filter"
         assert ref is not None, "Invalid reference value"
-        assert abs(ref[-3] - actual[-3]) < 0.25, "Luma (Y) out of baseline range"
+        # Only assert if actual Luma (Y) is lower than reference, using specified range condition
+        if actual[-3] < ref[-3]:
+          # Ensure the difference is within the acceptable range [0, 0.25)
+          assert 0 <= ref[-3] - actual[-3] < 0.25, f"Luma (Y) is lower than reference value: {actual[-3]} < {ref[-3]}"
```
uartie, line 138: If luma is 100 in non-passthrough sharpen level, then it's a bug and this will cause that bug to escape.

uartie: This change will not capture the issue reported in https://github.com/intel/media-driver/issues/1881

PR #703, opened 2024-11-22
Revert "Enhance PSNR Check for Luma (Y) Channel in VPP Sharpen Filter"

uartie:
Reverts intel/vaapi-fits#697


uartie:
> Reverts #697

See https://github.com/intel/vaapi-fits/pull/697#discussion_r1854027981

PR #705, opened 2025-03-27
framereader: fix read_frame_I010

gitgjogh:
as va.h defined

/** I010: three-plane 10-bit YUV 4:2:0.
 *
 * Each sample is a two-byte little-endian value with the top six bits ignored.
 * The three planes contain Y, V and U respectively.
 */
#define VA_FOURCC_I010          0x30313049


gitgjogh:
hi @uartie @FocusLuo @Hangjie22Coder , please take review

lib/framereader.py
```diff
@@ -96,9 +96,9 @@ def read_frame_I010(fd, width, height):
   size    = width * height
   size2   = width2 * height2
 
-  y = numpy.fromfile(fd, dtype=numpy.uint16, count=size).reshape((height, width)) >> 6
-  u = numpy.fromfile(fd, dtype=numpy.uint16, count=size2).reshape((height2, width2)) >> 6
-  v = numpy.fromfile(fd, dtype=numpy.uint16, count=size2).reshape((height2, width2)) >> 6
+  y = numpy.fromfile(fd, dtype=numpy.uint16, count=size).reshape((height, width)) & 0x03ff
+  u = numpy.fromfile(fd, dtype=numpy.uint16, count=size2).reshape((height2, width2)) & 0x03ff
+  v = numpy.fromfile(fd, dtype=numpy.uint16, count=size2).reshape((height2, width2)) & 0x03ff
```
uartie, line 101: This has the same result as `>> 6`.

uartie (changes requested):
This change doesn't do anything new/different.

lib/framereader.py
uartie: My bad.  I stand corrected.  I should've observed it closer ;).

It looks like we're reading the u and v in the wrong order, too!

* The three planes contain Y, V and U respectively.

Can you fix that, also?

uartie:
Please create a unit test, too, in test/self/formats.py

uartie:
... and please add more details to the actual git commit message body.

gitgjogh:
> Please create a unit test, too, in test/self/formats.py

I don't quite sure how you generate the assets. Never mind. I write a small tools to convert P010 to I010.

Seem test/self/formats.py just make read test without content check, I made inverse  I010 to P010 convertion to assert right.

lib/framereader.py
gitgjogh: @uartie done

uartie (approved):

uartie:
@Bin-CI please fix CI or delegate to someone who can.

